PR #5, opened 2019-10-31
Rename fetch to @adeira/fetch

tbergq:


src/fetch/package.json
```diff
@@ -1,12 +1,12 @@
 {
-  "name": "@kiwicom/fetch",
+  "name": "@adeira/fetch",
   "private": false,
   "license": "MIT",
   "version": "2.5.1",
```
mrtnzlml: We should reset the version and just backup the changelog like I did here: https://github.com/adeira/universe/blob/master/src/eslint-config-kiwicom/CHANGELOG.md (no new changelog entry - it's new package)

scripts/publishNPMPackages.js
```diff
@@ -24,7 +24,7 @@ import { invariant } from '@kiwicom/js';
       '@kiwicom/babel-preset-kiwicom',
       '@kiwicom/eslint-config',
       '@kiwicom/eslint-config-nitro',
-      '@kiwicom/fetch',
+      '@adeira/fetch',
```
mrtnzlml: Could you please remove everything else from here (and keep only this line)? It won't be 1:1 since not everything is necessary to publish now.

src/fetch/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Rename package to `@adeira/fetch`
```
mrtnzlml: Pls remove this line and just backup this file (see Eslint Config changelog for an example).

src/fetch/package.json
tbergq: Sure, I was thinking about that as well, I will just set it to `1.0.0` then

tbergq:
@mrtnzlml please check again

PR #14, opened 2019-11-01
Make docs work again, remove blog

mrtnzlml:


jaroslav-kubicek:
I would change also algolia settings to use different client https://github.com/adeira/universe/pull/14/files#diff-ca2cac1af1f122a3d5b210718b6f5a85R74

mrtnzlml:
@jaroslav-kubicek How should I update it? Should we probably remove the whole Algolia for now? They could provide us the whole search when we are done with the changes for free (see my website).

mrtnzlml:
@jaroslav-kubicek OK changed, please have a look.

jaroslav-kubicek:
rebase is needed @mrtnzlml

PR #12, opened 2019-10-31
CI: add action for NPM publisher

mrtnzlml:
Closes: https://github.com/adeira/universe/issues/11

PR #20, opened 2019-11-02
Rename relay to @adeira/relay

### LINKED ISSUE #10: Rename @adeira/relay

tbergq:
closes #10

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -36,7 +36,7 @@ module.exports = {
           },
           {
             name: 'react-relay',
-            message: 'Use @kiwicom/relay instead.',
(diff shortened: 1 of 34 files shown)


mrtnzlml:
@tbergq Please check the conflicts. You could probably skip the `scripts/publishNPMPackages.js` changes in these PRs since it doesn't work yet.

tbergq:
`publishNPMPackages` was not the only conflict, and we probably want to use this file soon, right? So I think it is ok to add the packages there 😊 

Please check again

mrtnzlml:
@tbergq See: #12 (we should publish only what's necessary - which Relay is, yes)

PR #24, opened 2019-11-03
Rename @adeira/example-relay

mrtnzlml:


mrtnzlml:
@tbergq Is it somehow possible to allow rebasing for the "update branch" instead of merging master into this branch?

tbergq:
> Is it somehow possible to allow rebasing for the "update branch" instead of merging master into this branch?

Not as far as I know, no 😞

PR #34, opened 2019-11-05
ESLint plugin: rename to "graphql-fragments"

jaroslav-kubicek:


michalsanger:
What's the motivation for this? It's usable also for use-cases without Relay?

jaroslav-kubicek:
@michalsanger IMHO yes, we don't do any relay-specific thing in the plugin and I don't plan to. I would like to emphasize that fragments are really part of GraphQL and should be used whatever your client is.

PR #38, opened 2019-11-06
Batch Adeira rename

mrtnzlml:
Still some places to go.


src/docs/docs/guides/eslint-config-migration.md
```diff
@@ -73,7 +73,7 @@ It could help to add info about warning and errors into issue description:
 ```text
 'jest/no-test-return-statement': OFF, // 1 error
 'jest/prefer-called-with': OFF, // 2 warnings
-'kiwicom-incubator/no-invalid-flow-annotations': OFF, // 44 errors
+'adeira-incubator/no-invalid-flow-annotations': OFF, // 44 errors
```
jaroslav-kubicek: `incubator` is probably also obsolete

mrtnzlml: All the plugins should be refactored into one or two - their names are taken anyway by the previous versions.

PR #40, opened 2019-11-06
Dependencies: upgrade flow-bin to the latest version 0.111.3

mrtnzlml:


src/relay/package.json
```diff
@@ -37,7 +37,7 @@
     "@adeira/test-utils": "^0.1.0",
     "babel-eslint": "^10.0.3",
     "eslint": "^6.5.1",
-    "flow-bin": "^0.111.0",
+    "flow-bin": "^0.111.3",
```
jaroslav-kubicek: do we need flow-bin here?

mrtnzlml: Yes, currently we do. It would be nice to clean it later though.

PR #43, opened 2019-11-06
Fix Relay Example

mrtnzlml:


src/example-relay/next.config.js
```diff
@@ -11,6 +11,6 @@ module.exports = withCustomBabelConfigFile(
       // @x-shipit-enable: 'babel.config.js',
       '.babelrc', // @x-shipit-disable
     ),
-    transpileModules: ['@kiwicom'],
+    transpileModules: ['@adeira'],
```
mrtnzlml: @tbergq This was the actual problem.

tbergq: Yeah, that would explain it 🙈

PR #44, opened 2019-11-07
Fix readInline flowtype

### LINKED ISSUE #37: Fix flowtypes for readInlineData
Current function looks like: 

```
export default function readInlineData(
  fragment: GraphQLTaggedNode,
  fragmentRef: GraphQLTaggedNode,
) {
  return _readInlineData(fragment, fragmentRef);
}
```
However, the fragmentRef is not a graphql tagged node. From the generated type, it looks like 

```
export type myInlineFunction = {|
  +name: ?{|
    +first: ?string,
    +last: ?string,
  |},
  +$refType: myInlineFunction$ref,
|};
export type myInlineFunction$data = myInlineFunction;
export type myInlineFunction$key = {
  +$data?: myInlineFunction$data,
  +$fragmentRefs: myInlineFunction$ref,
};
```

Where the fragmentRefType should be `myInlineFunction$key`

tbergq:
closes #37


src/relay/src/readInlineData.js
```diff
@@ -4,9 +4,12 @@ import { readInlineData as _readInlineData } from 'react-relay';
...
-  fragment: GraphQLTaggedNode,
-  fragmentRef: GraphQLTaggedNode,
-) {
+opaque type FragmentReference = empty;
+
+export default function readInlineData<
+  TRef: FragmentReference,
+  TData,
+  TKey: ?{ +$data?: TData, +$fragmentRefs: TRef, ... },
+>(fragment: GraphQLTaggedNode, fragmentRef: TKey): ?TData {
```
mrtnzlml: Seems like `TData` should be nullable only when `fragmentRef` is nullable, see: https://github.com/facebook/relay/blob/bb33efe09529e231ecccdbaaf96cb0fd69f86301/packages/relay-runtime/store/readInlineData.js#L25-L40

I will merge it so we can iterate.

tbergq: Yeah, I chose to go with the latter, ok, let's iterate on it.

mrtnzlml: @tbergq It would be better to go with both. It's like a variant. Similarly for example `JSON.stringify` [is also defined twice](https://github.com/facebook/flow/blob/f2a168ebbc71c6bc51c8565768b11c6432a8d216/lib/core.js#L511-L520) since it's more complicated in reality.

PR #46, opened 2019-11-07
Release @adeira/logger 0.1.0

tbergq:


tbergq:
done 🚀

tbergq:
> You should also flip the "private": true switch in package.json.

Should we have some scanner test for this?

mrtnzlml:
That would be nice but I am not sure how to do it exactly. The NPM publisher needs some improvements in this area. I would be happy if you put it on your todo list though and make some small improvements when you go around. :)

PR #84, opened 2019-11-08
Test that npm packages has private false

tbergq:


mrtnzlml:
Hi @tbergq, sorry for the delays. I wanted to check one thing more closely, but I didn't get to it yet.

tbergq:
No worries, take your time. Note that packages are now alphabetically sorted 🤓

tbergq:
This script will also fail if we try to release a package that does not exist. Which is how I found #103

src/monorepo-utils/src/Workspaces.js
```diff
@@ -69,4 +70,17 @@ module.exports = {
     });
     return packageJSONLocations;
   },
+
+  getWorkspaceLocationMap(): $ReadOnlyMap<string, string> {
```
mrtnzlml: @tbergq Why isn't the `getWorkspacesSync` good enough for your use-case? The functions which are already available here seems to be covering your case as well. 🤔

scripts/publishedPackages.json
```diff
@@ -0,0 +1,14 @@
+[
```
mrtnzlml: TODO before merging: rebase this PR and update this list.

src/monorepo-utils/src/Workspaces.js
tbergq: `getWorkspacesSync ` only gives me the path to each package.json file. I would have to read every json file and check the name.

I am all for using existing functionality, I just can't find anything that fits the use case well

tbergq: @mrtnzlml please check again

src/monorepo-scanner/src/scans/NpmPackages.scan.js
```diff
@@ -0,0 +1,25 @@
...
+  workspaceMap.set(packageJson.name, packageJson);
+}
+
+// $FlowAllowDynamicImport
+const npmPackages = require(path.join(root, 'scripts', 'publishedPackages.json'));
+
+for (const npmPackage of npmPackages) {
+  test(`${npmPackage} has private false flag`, () => {
+    const packageJson = workspaceMap.get(npmPackage);
+    expect(packageJson != null).toGiveHelp('You are trying to release a non existing package');
```
jaroslav-kubicek: `!=` instead `!==` is intentional?

tbergq: Probably it will only be undefined here, so I could change that yes

tbergq: 🚀

src/monorepo-scanner/src/scans/NpmPackages.scan.js
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// $FlowAllowDynamicImport
+const npmPackages = require(path.join(root, 'scripts', 'publishedPackages.json'));
+
+for (const npmPackage of npmPackages) {
+  test(`${npmPackage} has private false flag`, () => {
+    const packageJson = workspaceMap.get(npmPackage);
+    expect(packageJson !== undefined).toGiveHelp(
+      'You are trying to release a non existing package',
```
mrtnzlml: Yaay! 👍

PR #93, opened 2019-11-09
Release eslint-plugin-adeira

tbergq:
These packages are needed by `@adeira/eslint-config`

Release `@adeira/eslint-config` version `0.2.0`

### DIFF
--- a/scripts/publishNPMPackages.js
+++ b/scripts/publishNPMPackages.js
@@ -33,6 +33,7 @@ import { invariant } from '@adeira/js';
       '@adeira/eslint-config',
       '@adeira/graphql-global-id',
       '@adeira/graphql-bc-checker',
+      'eslint-plugin-adeira',
     ]),
   });
 })();
--- a/src/eslint-config-adeira/CHANGELOG.md
+++ b/src/eslint-config-adeira/CHANGELOG.md
@@ -1,7 +1,9 @@
 # Unreleased
 
+# 0.2.0
+
 - Remove `eslint-plugin-react-native`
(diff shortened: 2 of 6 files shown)


scripts/publishNPMPackages.js
```diff
@@ -31,6 +31,8 @@ import { invariant } from '@adeira/js';
       '@adeira/test-utils',
       '@adeira/flow-bin',
       '@adeira/eslint-config',
+      'eslint-plugin-adeira',
+      'eslint-plugin-relay-imports',
```
mrtnzlml: I am not sure if this is a good idea since it was used before we migrated and therefore we might be messing with the versions and unexpectedly release something they don't have under control. How about moving this code under `eslint-plugin-adeira`?

tbergq: Ok, I will see if I can move it under `eslint-plugin-adeira`.

mrtnzlml:
I assume this can be closed after https://github.com/adeira/universe/pull/95, right? @tbergq

tbergq:
We still need to release eslint-plugin-adeira. I will rebase this tomorrow

src/eslint-config-adeira/CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 # Unreleased
 
+# 0.2.0
+
 - Remove `eslint-plugin-react-native`
-- Move rules from `eslint-plugin-relay-imports` under this package
```
tbergq: This was never under this packages 🙈

tbergq:
@mrtnzlml please have another look

Issue #91, opened 2019-11-09
New eslint rule 'no-exact-indexer'


mrtnzlml:
To disallow ~totally~ broken type like this one: `type X = {| +[key: string]: any |}` 🙈

mrtnzlml:
I have to postpone this, because, turned out it's maybe better to leave it half functional rather than completely disallow some use-cases. See:

```js
type S = {| [key: string]: string |};

const s: S = { key: 'value' }; // Error: Cannot assign object literal to `s` because property `key` is missing in `S` but exists in object literal.

function test(x: S): string {
  x.y = 'string'; // OK
  return x.test; // OK
}
```

[flow.org/try](https://flow.org/try/#0C4TwDgpgBAylC8UDeAfKBtA1hEAuKAzsAE4CWAdgOYC6+RZVUKAvgNwBQ7AxgPblGF8cREijY8UAOQA3AIYAbAK4RJUNlAD0GqAFFixHsU4AzReS7BSfKMAhEAFAA9cMAJR0SFSsnZQojgDoQBCl6L0lWTW0AeQBpXyhiCGBFYnJ-ANsiSK0oOPZmdiA)

This pattern is especially handy if you want to annotate "any arbitrary *exact* object" and work with the properties. The only known caveat is that you cannot completely rewrite it (but you can access the properties individually and it does the typechecking correctly).

PR #94, opened 2019-11-09
Flow: disallow exact objects with indexer properties

mrtnzlml:
Related issue: https://github.com/facebook/flow/issues/3162

### DIFF
--- a/src/example-relay/src/Hotels/SSRQueryRenderer.js
+++ b/src/example-relay/src/Hotels/SSRQueryRenderer.js
@@ -5,7 +5,7 @@ import { QueryRenderer, type GraphQLTaggedNode } from '@adeira/relay';
 
 import createRelayEnvironment from '../createRelayEnvironment';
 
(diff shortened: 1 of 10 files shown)


mrtnzlml:
Draft, see the explanation why in #91

src/example-relay/src/Hotels/SSRQueryRenderer.js
```diff
@@ -5,7 +5,7 @@ import { QueryRenderer, type GraphQLTaggedNode } from '@adeira/relay';
 
 import createRelayEnvironment from '../createRelayEnvironment';
 
-type RendererProps = {| +[key: string]: any |};
+type RendererProps = any; // It should be any arbitrary *exact* object but how to annotate it?
```
mrtnzlml: Relay also annotates this with `any`, see: https://github.com/facebook/relay/blob/master/packages/react-relay/ReactRelayQueryRenderer.js#L47

PR #95, opened 2019-11-10
Move relay-import rules under eslint-plugin-adeira

tbergq:
Also delete unused react-native-eslint plugin


src/eslint-plugin-adeira/README.md
```diff
@@ -1,20 +1,20 @@
...
 
 # Usage
 
-Add `kiwicom-incubator` to the plugins section of your `.eslintrc` configuration file:
+Add `adeira` to the plugins section of your `.eslintrc` configuration file:
 
 ```json
 {
-  "plugins": ["eslint-plugin-kiwicom-incubator"]
+  "plugins": ["eslint-plugin-eslint-plugin-adeira"]
```
mrtnzlml: This seems to be wrong (duplicated `eslint-plugin-`).

src/eslint-config-adeira/ourRules.js
```diff
@@ -635,12 +626,10 @@ module.exports = {
...
-  'relay-imports/no-values': ERROR,
-  'relay-imports/type-must-exist': ERROR,
-
-  // Kiwi.com custom rules
+  // Adeira custom rules
   'adeira/only-nullable-fields': ERROR,
   'adeira/no-invalid-flow-annotations': ERROR,
   'adeira/no-internal-flow-type': NEXT_VERSION_ERROR, // TODO: ERROR
+  'adeira/relay-import-no-values': ERROR,
+  'adeira/relay-import-type-must-exist': ERROR,
```
mrtnzlml: Should we keep the plural version? 🤔 Original rules were `relay-imports` with "s".

src/eslint-plugin-adeira/README.md
tbergq: Probably from search and replace, yes, I will fix it.

src/eslint-config-adeira/ourRules.js
tbergq: My idea was that before, it would have been `relay-imports/no-values`, but now it is part of `adeira/` so singular sounded better in my head. 

Not really that important to me. If you want me to change it back, I am fine with it it.

PR #103, opened 2019-11-12
Name graphql-global-id package correctly npm relase file

tbergq:


mrtnzlml:
@tbergq Would it be possible to test for this as well? Or should we wait for #84? (I am delaying it, sorry 🙈)

tbergq:
#84 Will also validate that the package exists, since scanner will fail trying to read non existing workspace.

PR #116, opened 2019-11-16
Add options for valid-test-folder rule

tbergq:


src/eslint-plugin-adeira/src/rules/__tests__/valid-test-folder.test.js
```diff
@@ -22,6 +22,11 @@ ruleTester.run('valid-test-folder', rule, {
       code: "it('works', () => {expect(true).toBe(true)})",
       filename: 'src/packages/test/__tests__/Module.spec.js',
     },
+    {
+      code: "it('works', () => {expect(true).toBe(true)})",
+      filename: 'src/packages/test/__tests__/Module.custom.jsx',
+      settings: { isTestRegex: '.custom.jsx$' },
```
mrtnzlml: I find the naming a bit strange. Why is it starting with "is" and not simply "testRegex" to follow the Jest settings? See: https://jestjs.io/docs/en/configuration#testregex-string-array-string (this also implies only one option is needed)

tbergq: We can remove `is`. But we would need to verify that it is actually a test first I think. If not, we could end up giving errors on non-test files?

mrtnzlml: I see, I didn't pay much attention to the implementation. So this regex is for iterating only test files, correct? Could you maybe explain it better in the docs?

tbergq: Yes, I can do that.

tbergq: I added some more explanation in the docs. Please check again.

Issue #83, opened 2019-11-08
Cleanup and move 'src/docs/docs/monorepo'


mrtnzlml:
Probably somewhere between README and CONTRIBUTING. Some files can be deleted or rewritten (since they are going to be outdated).

Issue #81, opened 2019-11-08
Cleanup and move 'src/docs/docs/guides'


mrtnzlml:
Probably to adeira.dev?

jaroslav-kubicek:
closed by #126

PR #126, opened 2019-11-16
Docs: cleanup old docs for preparation to publish adeira.dev

jaroslav-kubicek:
Closes https://github.com/adeira/universe/issues/83 & https://github.com/adeira/universe/issues/81

I removed all remaining docs, feel free to suggest what should be kept.

Screenshot 2019-11-17 00 55 31


tbergq:
> I removed all remaining docs, feel free to suggest what should be kept.

If there is anything we need at a later stage, we can always recover it from the git log.

mrtnzlml:
@jaroslav-kubicek Where is `src/docs/static/img/logo-no-text.png` coming from?

PR #131, opened 2019-11-17
Docs: fix link on welcome page

jaroslav-kubicek:


src/docs/siteConfig.js
```diff
@@ -11,7 +13,7 @@ module.exports = {
   editUrl: 'https://github.com/adeira/universe/tree/master/src/docs/docs/',
   cname: 'adeira.dev',
 
-  customDocsPath: 'docs/docs',
+  customDocsPath: `${path.basename(__dirname)}/docs`,
```
mrtnzlml: We should migrate to Docusaurus 2 where this works a bit differently and I'd say better. :)

PR #138, opened 2019-11-19
Relay: Use console.warn in createRequestHandler

tbergq:


mrtnzlml:
@tbergq What kind of crash is that? Almost sounds like we should rather use logger to fix it there instead of changing it like this.

tbergq:
console.error in RN automatically triggeres `red screen of death`, as console.warn triggers yellow boxes
Simulator Screen Shot - iPhone 11 - 2019-11-20 at 11 57 30

Not sure if logger will fix it also uses console.error underneath. But probably it could be used 🤷‍♂

PR #142, opened 2019-11-20
Upgrade dependencies between packages

jaroslav-kubicek:
This could happen automatically... 🤔


tbergq:
> This could happen automatically... 

There could be some script triggered manually, or maybe a scanner that tests that we upgrade our internal packages.

PR #171, opened 2019-11-23
Dependabot: enable 'automerged_updates'

mrtnzlml:


jaroslav-kubicek:
should we also revisit peer dependencies from time to time? these usually stay untouched

PR #173, opened 2019-11-24
Typescript support for graphql-global-id

michalsanger:
First steps forward TS support. I just hope NPM publisher includes `.d.ts` file automatically :-)


tbergq:
> First steps forward TS support. I just hope NPM publisher includes .d.ts file automatically :-)

I think it should yes, the only thing ignored is __tests__. You can do `(cd src/graphql-global-id/ && npm pack --dry-run)` and you will see that `src/GlobalID.d.ts` is included

PR #174, opened 2019-11-24
eslint-config: Lint all when .eslintignore changes

tbergq:


src/eslint-config-adeira/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Unreleased
 
 
+# 0.4.0
+
+- Lint all files when `.eslintignore` changes
```
mrtnzlml: We shouldn't write a changelog for versions before 1.0 - they can all be breaking by design and the scanner ignores most of the checks for such versions.

src/eslint-config-adeira/runner/__tests__/shouldLintAll.test.js
```diff
@@ -13,6 +13,8 @@ test.each([
   'package.json',
   'src/packages/relay/package.json',
   '/Users/code/universe/src/packages/relay/package.json',
+  '.eslintignore',
+  '/Users/code/universe/src/packages/relay/.eslintignore',
```
mrtnzlml: Eslint ignore file which is not in the root doesn't work if I recall correctly.

src/eslint-config-adeira/runner/shouldLintAll.js
```diff
@@ -3,5 +3,5 @@
 export default function shouldLintAll(filename: string): boolean %checks {
   // Eslint configs can be nested (not only the root path).
   // See: https://eslint.org/docs/user-guide/configuring#configuration-file-formats
-  return /\.eslintrc(?:\.(?:js(?:on)?|ya?ml))?$|package.json$/.test(filename);
+  return /\.eslintrc(?:\.(?:js(?:on)?|ya?ml))?$|package.json$|.eslintignore$/.test(filename);
```
mrtnzlml: I have this feeling that we should fix this regexp to ignore dots. Something like this:

```diff
-   return /\.eslintrc(?:\.(?:js(?:on)?|ya?ml))?$|package.json$|.eslintignore$/.test(filename);
+   return /\.eslintrc(?:\.(?:js(?:on)?|ya?ml))?$|package\.json$|\.eslintignore$/.test(filename);
//                                                        ^       ^
```

src/eslint-config-adeira/runner/__tests__/shouldLintAll.test.js
tbergq: Ahh, yes, you are right

src/eslint-config-adeira/runner/shouldLintAll.js
tbergq: You are right 🙈 Will do it right away

tbergq: #178

PR #175, opened 2019-11-24
Relay: expose language plugins interface

mrtnzlml:
This exposes public interface for changing the language plugin. Currently, people are mostly interested in 'javascript' or 'typescript' languages so I added support only for these two. Javascript is and will stay as a default.

### DIFF
--- a/src/relay/package.json
+++ b/src/relay/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@adeira/relay",
   "private": false,
-  "version": "0.3.0",
+  "version": "0.4.0",
   "main": "src/index",
   "module": false,
   "sideEffects": false,
@@ -28,6 +28,7 @@
     "is-ci": "^2.0.0",
     "react-relay": "^7.1.0",
     "relay-compiler": "^7.1.0",
+    "relay-compiler-language-typescript": "^10.1.1",
     "relay-config": "^7.1.0",
     "relay-runtime": "^7.1.0"
   },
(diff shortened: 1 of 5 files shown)


src/relay/src/compiler/index.js
```diff
@@ -23,6 +23,7 @@ type ExternalOptions = {|
   +exclude: $ReadOnlyArray<string>,
   +validate: boolean,
   +watch: boolean,
+  +language: 'javascript' | 'typescript', // TODO: detect this automatically based on tsconfig.json file?
```
jaroslav-kubicek: I could imagine codebase with both...

src/relay/package.json
```diff
@@ -28,6 +28,7 @@
     "is-ci": "^2.0.0",
     "react-relay": "^7.1.0",
     "relay-compiler": "^7.1.0",
+    "relay-compiler-language-typescript": "^10.1.1",
```
jaroslav-kubicek: does it have to be dependency? `import(optionFromCli)` is no?

src/relay/src/compiler/index.js
mrtnzlml: This is currently impossible to do with Relay Compiler as far as I know.

src/relay/package.json
mrtnzlml: It doesn't have to be dependency but you'd have to require it manually when you need it. I added it here to simplify the process since it doesn't increase the bundle size or anything (it's a development thingy only for the Relay Compiler). We can tune it later when there is a real concern not to do it.

src/relay/src/compiler/index.js
jaroslav-kubicek: I meant you could have both flowconfig and tsconfig.json in repo, because you previously used flow and then switched to typescript, just wild idea what's possible... :)

mrtnzlml: OK, good point - no automatic detection then.

PR #176, opened 2019-11-24
Relay CLI: remove include/exclude options

mrtnzlml:
These options were added because of one legacy project which we don't have under control anymore. It was to enable multiple schemas in the same repository which is highly discouraged.


src/relay/src/compiler/index.js
```diff
@@ -20,7 +20,6 @@ type ExternalOptions = {|
   +schema: string,
   +persistMode: 'fs', // TODO consider more generic: +persistFunction?: ?(query: string) => Promise<string>,
   +include: $ReadOnlyArray<string>,
```
mrtnzlml: Wait, this should be removed as well... 😬

PR #181, opened 2019-11-24
example-relay: Fix jest unit tests in shipped version of the example

jaroslav-kubicek:
Problem:

- clone https://github.com/adeira/relay-example
- `yarn`
- `yarn flow` -> flow fails on unit tests because `define` and `it` is unkown global variable

There should be 2 remaining flow issues in https://github.com/adeira/relay-example after this MR

should be `src/example-relay/scripts` also stripped? 🤔


.eslintignore
```diff
@@ -5,3 +5,6 @@ node_modules/
 
 # Reported issue (must be excluded here since it fails on parsing): https://github.com/babel/babel-eslint/pull/777
 src/babel-preset-adeira/src/__tests__/__fixtures__/features/internal-slot-properties.js
+
+# flow-typed for shipit
```
jaroslav-kubicek: is there better way?

src/example-relay/__github__/jest.config.js
```diff
@@ -0,0 +1,6 @@
+// @flow strict
+
+module.exports = {
+  rootDir: __dirname,
+  testMatch: ['<rootDir>/**/__tests__/**/?(*.)+(spec|test).js?(x)'],
+};
```
mrtnzlml: Why do you need to replace this file? Could we just modify the currently existing one?

.eslintignore
mrtnzlml: Nothing I am aware of.

mrtnzlml: I'd remove the comment though. It's not for Shipit. It's part of the project code.

src/example-relay/__github__/jest.config.js
jaroslav-kubicek: original one uses `src/example-relay/scripts/jest/custom-transformer.js` -> shouldn't we strip this custom transformer as well?

jaroslav-kubicek:
```
// This configuration is merged with the project configuration defined
// in this monorepo root.
module.exports = {
  rootDir: __dirname,
  transform: {
    '^.+\\.js$': '<rootDir>/scripts/jest/custom-transformer.js',
  },
};
```

vs 

```
module.exports = {
  rootDir: __dirname,
  testMatch: ['<rootDir>/**/__tests__/**/?(*.)+(spec|test).js?(x)'],
};
```

mrtnzlml: OK, I will leave it up to you. It could be easier to leave there even the transformer if it doesn't break anything for OSS.

jaroslav-kubicek: thanks, it breaks because it tries to find `.babelrc.js` which is gone

mrtnzlml: This can actually be easily fixed - check Shipit docs and search for `@x-shipit-enable` comments.

mrtnzlml: That's also a way how to enable/disable some lines for `jest.config.js` (as an alternative approach).

jaroslav-kubicek: hmm, ok, I considered two options

- having one jest file an use `@x-shipit-enable`
- having two files for jest

I went for second one because this transform is not needed and require in that file produces another flow issue which would require me to put additional suppress comment into flowconfig

src/example-relay/flow-typed/globals.js
```diff
@@ -0,0 +1 @@
+declare var __DEV__: boolean;
```
mrtnzlml: Why aren't these `flow-typed` files under `__github__` as well? (they are used in `src/example-relay/__github__/.flowconfig` for example so it would make sense)

jaroslav-kubicek: fixed

PR #191, opened 2019-11-28
Add option to pass logger when creating environment

tbergq:
Especially when debugging things on the server, the logging of the fragments and response is just noise. It would be nice to easily deactivate the logger in these cases by passing null.

### DIFF
--- a/src/example-relay/next.config.js
+++ b/src/example-relay/next.config.js
@@ -11,6 +11,7 @@ module.exports = withCustomBabelConfigFile(
       // @x-shipit-enable: 'babel.config.js',
       '.babelrc', // @x-shipit-disable
     ),
-    transpileModules: ['@adeira'],
+    // @kiwicom/orbit-components is using `@adeira/js`, and needs to be  tranpiled as well because of this
+    transpileModules: ['@adeira', '@kiwicom'],
   }),
 );
--- a/src/example-relay/package.json
+++ b/src/example-relay/package.json
@@ -5,7 +5,7 @@
   "private": true,
   "dependencies": {
     "@adeira/fetch": "^1.0.0",
-    "@adeira/relay": "^0.5.0",
+    "@adeira/relay": "^0.6.0",
     "@kiwicom/orbit-components": "^0.68.0",
     "date-fns": "^2.8.1",
     "next": "^9.1.4",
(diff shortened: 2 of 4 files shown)


mrtnzlml:
Cou you instead reduce the noise or somehow modify the logger, please?

src/example-relay/next.config.js
```diff
@@ -11,6 +11,7 @@ module.exports = withCustomBabelConfigFile(
       // @x-shipit-enable: 'babel.config.js',
       '.babelrc', // @x-shipit-disable
     ),
-    transpileModules: ['@adeira'],
+    // @kiwicom/orbit-components is using `@adeira/js`, and needs to be  tranpiled as well because of this
+    transpileModules: ['@adeira', '@kiwicom'],
```
mrtnzlml: I don't understand this. Could you explain it better what's the problem? This should not be the case.

tbergq:
The logger is fine, it is just that when you want to focus on other output in you terminal happening on the server, you might miss it, if the fragment is long, and response is long. 

You cannot collapse it in the terminal, as you can in the browser.

src/example-relay/next.config.js
tbergq: in `@kiwicom/orbit-components`, there is `var _js = require('@adeira/js')`; Then when starting example relay, we get error because `export is undefined`. It is mapping this to our un-transpiled source code.

mrtnzlml:
> You cannot collapse it in the terminal, as you can in the browser.

Yes, that's why I wanted to stop printing the queries on the server instead of removing it completely. Just an alternative output for CLI. What do you think?

src/example-relay/next.config.js
mrtnzlml: So they are bundling untranspiled `adeira/js`? 🤔 How? I still don't understand. I guess I will have to check it closely.

tbergq: Try starting `example relay` locally.

> So they are bundling untranspiled adeira/js?

I don't think that is how it works. They transpile their code own code, and we download their dependencies. Then our repo will not download `@adeira/js` since it is an internal workspace. And requiring it from `@kiwicom/orbit` will point it to our untranspiled workspace

tbergq:
> Yes, that's why I wanted to stop printing the queries on the server instead of removing it completely.

Ok, that will work as well.

tbergq:
I will create a new PR for this

PR #195, opened 2019-11-29
Add is browser function

tbergq:


tbergq:
I plan to use this function to determine if we should log with relay logger on server.

src/js/src/isBrowser.js
```diff
@@ -0,0 +1,10 @@
+// @flow strict
+
+// https://stackoverflow.com/questions/17575790/environment-detection-node-js-or-browser/31090240#31090240
+// eslint-disable-next-line no-new-func
+const testIsBrowser = new Function('try {return this===window;}catch(e){ return false;}');
+
+export default function isBrowser(): boolean {
+  // We need to add !! to be able to type the return type as boolean
+  return !!testIsBrowser();
```
jaroslav-kubicek: `testIsBrowser() === true` not possible?

tbergq: Probably yes, is it somehow better?

jaroslav-kubicek: probably not, comment should not be needed in that case :)

tbergq: I feel it would be necessary anyway. I need to explain why we simply don't do `return testIsBrowser()`, and that doesn't change with `return testIsBrowser() === true`

Issue #189, opened 2019-11-27
@adeira/relay: fetch-schema script should support .babelrc


jaroslav-kubicek:
When running adeira-fetch-schema on repository with `.babelrc` only, it fails to find babel config: 

```
adeira-fetch-schema --resource=https://graphql.kiwi.com
/Users/jaroslavkubicek/projects/nitrolib/node_modules/@babel/core/lib/config/partial.js:50
        throw Object.assign(new Error(`Babel was run with rootMode:"upward" but a root could not ` + `be found when searching upward from "${rootDir}"`), {
        ^

Error: Babel was run with rootMode:"upward" but a root could not be found when searching upward from "/Users/jaroslavkubicek/projects/nitrolib"
```

mrtnzlml:
You should upgrade to Babel 7 properly by using project-wide configuration `babel.config.js`, see: https://babeljs.io/docs/en/config-files#project-wide-configuration

I encountered this issue as well on one project and I am not sure we should modify it in this package (it's simple, we'd have to use root mode `upward-optional`). It seems like a problem of the project where people ignore the new configuration system for Babel 7.

jaroslav-kubicek:
of course 🙂  possible solution can be also upgrading error message explaining this :)

mrtnzlml:
Sure, feel free to send a PR to Babel since this is nothing Universe specific. :)

jaroslav-kubicek:
Closed via https://github.com/babel/babel/pull/10778

mrtnzlml:
Very nice! 👍

mrtnzlml:
This will also be allowed from our side: https://github.com/adeira/universe/pull/203 (if it goes through the review)

It's because the code which is enforcing the "upward" mode is necessary only for Universe development and should not affect our NPM users. So I changed my mind and now I think this issue is very valid.

Thank you for your contribution into Babel itself though - that's really great no matter this change! :)

mrtnzlml:
Side note: I believe it's still valid to use project-wide config for Babel 7+ projects... :)

PR #203, opened 2019-12-02
Relay: do not enforce project wide configuration

mrtnzlml:
This feature was requested here https://github.com/adeira/universe/issues/189 and was greatly improved in Babel itself here https://github.com/babel/babel/pull/10778. It can however be annoying and sometimes difficult to do such modification so I changed my mind and I'd like to allow it after all. It's because this code where we require "upward" mode is necessary only for Universe development and should not affect users of our NPM package.


tbergq:
LGTM, but you have some merge conflicts here 😞

PR #204, opened 2019-12-02
Relay: Do not log on server

tbergq:


src/relay/src/RelayLogger.js
```diff
@@ -63,7 +65,7 @@ function logGroup(logEvent, groupBody) {
 
 // See: https://github.com/facebook/relay/commit/da9a57cb0b7ab9bedf82e3d1dddc17a0ad9e4d92
 export default function RelayLogger(logEvent: LogEvent) {
-  if (!__DEV__) {
+  if (!__DEV__ || !isBrowser()) {
```
mrtnzlml: Does it make sense to disable the logging completely? I imagined modifying the behavior of `execute.start` for example (so you still get some logs). But I am OK even with this solution - at least for now.

tbergq: Well, this was why my initial idea was to allow the user to pass in a function or null, so we have our default behaviour, and the possibility to configure it however the consumer of the package would like 🤷‍♂

mrtnzlml:
@tbergq You might need to bump the version again since 0.6 was released in a different PR.

tbergq:
@mrtnzlml should be good to go now

src/relay/src/RelayLogger.js
mrtnzlml: OK, understand. It would be nice to have some default working configuration which is not annoying and then make it configurable as well. This logger is still somehow experimental so I didn't really expect people would like to change it.

So, what do you propose? Should we go with this solution you just submitted? I think it's a good starter. In fact, maybe it's the ultimate solution since I barely check the server logs for Relay stuff.

mrtnzlml:
OK, let's push it this way. Do you think it would be complicated to test it as well?

tbergq:
No, it shouldn't be. Do you want me to add a test for it as well?

mrtnzlml:
next PR please if you have time and you know how, it's not that critical

src/relay/src/RelayLogger.js
tbergq: Yes, probably, you will get it in the console in the browser anyway. 

Probably this will stop logging on RN, so if someone needs it there, we will need to tune it I guess.

PR #226, opened 2019-12-09
relay: Add option for Gc release size in createEnvironment

tbergq:


src/relay/src/createRelayStore.js
```diff
@@ -4,9 +4,10 @@ import { RecordSource, Store } from 'relay-runtime';
 
 import type { RecordMap } from './runtimeTypes.flow';
 
-export default function createRelayStore(records: ?RecordMap) {
-  const source = new RecordSource(records); // TODO
-  return new Store(source, {
-    gcReleaseBufferSize: 0, // TODO: use this instead of our burst cache
```
mrtnzlml: Could we please leave the comment there as well as the zero default value?

tbergq: I can leave the comment, the default fallback is already 0.

https://github.com/facebook/relay/blob/4c932cdbb501ec0144e6d66acfbb3f005b843b57/packages/relay-runtime/store/RelayModernStore.js#L80

tbergq:
@mrtnzlml please check again

PR #231, opened 2019-12-11
shipit: Remove non-existing ".github" dir from "relay-example" config

jaroslav-kubicek:
Relay-example is not updated for last weeks because `.github` dir does not exist yet. 

We can always put it back but now it's better to keep repositories in sync.


mrtnzlml:
This should not be removed. This PR should be imported instead: https://github.com/adeira/relay-example/pull/6/files (I added it specifically because of this but didn't have time to import it yet)

The PR cannot be imported without this.

PR #235, opened 2019-12-13
graphql-global-id: toGlobalId function added

michalsanger:
Added `toGlobalId` to it's possible to create global ID not only in `GraphQLObjectType` field.

### DIFF
--- a/src/graphql-global-id/src/GlobalID.js
+++ b/src/graphql-global-id/src/GlobalID.js
@@ -32,6 +32,10 @@ export function fromGlobalId(opaqueID: string): string {
   return unbasedGlobalID.substring(delimiterPos + 1);
 }
 
+export function toGlobalId(type: string, id: string | number): string {
+  return base64(`${type}:${id}`);
+}
+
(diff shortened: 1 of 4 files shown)


src/graphql-global-id/src/GlobalID.js
```diff
@@ -32,6 +32,10 @@ export function fromGlobalId(opaqueID: string): string {
   return unbasedGlobalID.substring(delimiterPos + 1);
 }
 
+export function toGlobalId(type: string, id: string | number): string {
```
tbergq: Could you please explain why we need to export this function?

tbergq: I mean, what is the use case?

michalsanger: I want to add `nodeId` into our current REST API so I can call GraphQL query that needs the global ID.

tbergq: Aha, I see 🙂

PR #237, opened 2019-12-15
Relay get data from request

tbergq:


mrtnzlml:
@tbergq I was thinking about this change - how about moving it into some other package like `relay-utils` or similar and focusing `adeira/relay` on Relay-only things? (setup, types, default configs, some extra checks)

tbergq:
Sure, makes sense. We could do that.

mrtnzlml:
@tbergq Could you help me with this cleanup, please? :)

I'd like to refocus the package more on Relay again, however, it's not that easy to draw the line and to answer questions like "why is this in `adeira/relay` and this cannot be" 😬

tbergq:
Sure, yes, I can help. 

Yes, well, I guess we need to figure out this `line` 🤷‍♂ 

I can justify having the getRequest in some util, because I guess only a subset of users will need this function.

mrtnzlml:
For me, the line is something like: Is it in Relay itself? Does it configure Relay (changes defaults)? Does it add Flow types? If so then it should be there. Otherwise probably not. Not sure if we follow this line everywhere though.

PR #241, opened 2019-12-21
Upgrade relay to 8.0.0

tbergq:


tbergq:
Will the deployed version of relay-example break if we do not release a new version of adeira/relay?

mrtnzlml:
> Will the deployed version of relay-example break if we do not release a new version of adeira/relay?

How do you mean it? The Universe version will use master but the Shipit version will probably use NPM version.

tbergq:
🧐 yes, so the relay example will be deployed with v7.1.0 of all relay packages. I guess it will still work, since we were not affected by any BC

mrtnzlml:
Feel free to bump adeira/relay version as well so it will be the same everywhere.

PR #243, opened 2019-12-21
Allow --inspect and --inspect-brk in monorepo-test-runner

tbergq:


mrtnzlml:
@tbergq Would be interesting to do something like this: https://github.com/adeira/universe/blob/d201b0bf4dc62b1989ec3e9bafb98c01629bf96a/src/monorepo-utils/bin/monorepo-babel-node.js#L15

(it's in the same package)

tbergq:
That looks better, updated. Please check again @mrtnzlml

tbergq:
Well, it was still approved so 🙈

PR #245, opened 2019-12-21
Upgrade eslint

tbergq:


mrtnzlml:
Seems like Dependabot is sleeping. I woke it up.

PR #257, opened 2019-12-22
Relay: remove burst cache

mrtnzlml:
Seems like the same idea should be eventually replaced with `gcReleaseBufferSize` so I want to clean the package a little bit. Would be great to come closer to the original Relay API again.


src/relay/src/createRequestHandler.js
```diff
@@ -44,33 +37,12 @@ export default function createRequestHandler(
     uploadables: ?Uploadables,
   ) {
     return Observable.create((sink: Sink) => {
-      const queryID = requestNode.text;
```
tbergq: We probably don't need `Observable` anymore?

mrtnzlml: Why not? It's still handly for proper error propagation or things like `poll`.

tbergq: Why is Observable better in that case? I would expect it to work the same now, just removing the observable, and keeping `customFetcher(requestNode, variables, uploadables).then().catch()`

mrtnzlml: Sort of explained in person - better errors propagation (`sink.error`) and ability to use `poll` which wouldn't work otherwise.

PR #262, opened 2019-12-26
Eslint-config: Allow to disable automatic fixes

jaroslav-kubicek:
Reasoning:

I like the way how eslint runner is opinionated yet I would like to have an option to change default behavior e.g. when adopting this tool on a large codebase which is too broken.


jaroslav-kubicek:
READMe fixed, but I don't understand why CI fails: https://github.com/adeira/universe/pull/262/checks?check_run_id=369491049

mrtnzlml:
@jaroslav-kubicek I'd say the error is because you are creating `CLIEngine` for every single file.

jaroslav-kubicek:
hmm, I was examing this and looks like we need to get rid of `create-jest-runner` first 😞

PR #264, opened 2019-12-26
relay: Mock "console.groupCollapsed" in unit tests

jaroslav-kubicek:
do not accidentally output calls to `console.groupCollapsed` when running unit tests:

Screenshot 2019-12-26 22 57 28

### DIFF
--- a/src/relay/src/__tests__/RelayLogger.test.js
+++ b/src/relay/src/__tests__/RelayLogger.test.js
@@ -5,14 +5,17 @@
 
 import Logger from '../RelayLogger';
 
-let spy;
+let log;
+let groupCollapsed;
 
 beforeEach(() => {
-  spy = jest.spyOn(console, 'log').mockImplementation(jest.fn());
+  log = jest.spyOn(console, 'log').mockImplementation(jest.fn());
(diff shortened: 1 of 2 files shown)


src/relay/src/__tests__/RelayLogger.test.js
```diff
@@ -5,14 +5,17 @@
...
 beforeEach(() => {
-  spy = jest.spyOn(console, 'log').mockImplementation(jest.fn());
+  log = jest.spyOn(console, 'log').mockImplementation(jest.fn());
+  groupCollapsed = jest.spyOn(console, 'groupCollapsed').mockImplementation(jest.fn());
 });
 
 afterEach(() => {
-  spy.mockRestore();
+  log.mockRestore();
+  groupCollapsed.mockRestore();
```
mrtnzlml: Would it be possible to actually assert the logs instead? The initial idea of the restriction on `setupTests.js` was to consider even the logs as an application output and enforce tests for them (or remove them).

I am fine even with this though since it was silenced even before. I just wanted to point out what should be the correct fix and why is it like this.

PR #271, opened 2019-12-28
Add new pacakge @adeira/relay-utils

tbergq:


src/relay-utils/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@adeira/relay-utils",
+  "private": false,
+  "version": "0.1.0",
+  "main": "src/index",
+  "module": "src/index.mjs",
```
mrtnzlml: We don't write `module` manually in Universe - it's being created and added automatically when exporting to NPM.

src/relay-utils/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "sideEffects": false,
+  "homepage": "https://github.com/adeira/universe/tree/master/src/relay-utils",
+  "description": "Relay utils used at Adeira ecosystem",
+  "license": "MIT",
+  "dependencies": {
+    "react-relay": "0.0.0-experimental-5f1cb628",
+    "relay-runtime": "^8.0.0",
+    "@adeira/relay": "0.11.0"
```
mrtnzlml: Isn't `adeira/relay` sufficient enough? 🤔

tbergq: Yes, well, I need relay-runtime, but I don't need react-relay

src/relay-utils/package.json
tbergq: Should we have a scanner test for this?

src/relay-utils/package.json
mrtnzlml: Ah, you mean you are using it directly in the code, right? My thinking was that `adeira/relay` is requiring all these dependencies by default but your approach is better.

tbergq: > Ah, you mean you are using it directly in the code, right? 

Yes, I am importing from `relay-runtime`.

src/relay-utils/package.json
mrtnzlml: Well, technically there is nothing wrong about this. We allow writing such configuration in which case the module is not being generated automatically. But since you don't have this MJS file then it doesn't make much sense to write it here manually. So yes and no. 😬

src/relay-utils/package.json
mrtnzlml: @tbergq Shouldn't you specify these as a peer dependency?

tbergq: It could be that instead yes 🤔 👍

src/relay-utils/package.json
tbergq: So the test could be if the file specified in module does not exist, the test should give an error. 

I think it could be useful.

src/relay-utils/package.json
tbergq: Updated.

src/relay-utils/package.json
mrtnzlml: Yes, feel free to create it in a separated PR. :)

tbergq: Will do

src/relay-utils/src/getDataFromRequest.js
```diff
@@ -1,9 +1,7 @@
 // @flow
 
 import { getRequest, createOperationDescriptor } from 'relay-runtime';
```
mrtnzlml: @tbergq I was thinking. Maybe we should publish another package `@adeira/relay-runtime` solely with the additional Flow types and maybe with some other DX improvements we have? 🤔

mrtnzlml: Not sure if we have some added value over flow-typed when it comes to relay-runtime though.

tbergq: Yeah, if it is just flow-types, we can rather contribute to `flow-typed`

mrtnzlml: I was thinking more and it makes sense to create it. There are other things in runtime like the Store, our Environment factories, networking and similar.

jaroslav-kubicek:
I'm just curious why it couldn't be in `@adeira/relay`? 🙂

PR #281, opened 2020-01-02
WIP: relay-compiler: Add transform fn to warn against usage of deprecated fields

jaroslav-kubicek:
**WIP**

When running relay-compiler, warn when a deprecated field is used in a fragment.

By default, it should just log a warning into the console, e.g.

```
User.surname is deprecated (./path/to/component/User.js): Use "lastName" field instead
```


jaroslav-kubicek:
I'm closing this for now until I decide upon next steps

PR #292, opened 2020-01-08
Update peer dependency versions

jaroslav-kubicek:


mrtnzlml:
@jaroslav-kubicek What is the motivation behind this change? Peer dependency should in this context describe a minimal supported version basically. So it seems more correct to have it like this. Or am I missing something? 🤔

jaroslav-kubicek:
you are right, the question is whether it would work with these older dependencies - we have all up to date so we don't know.

mrtnzlml:
OK, I am fine with this argument. :)

PR #293, opened 2020-01-09
eslint-config: Add "--no-warnings" option to runner to suppress warnings

jaroslav-kubicek:
Motivation:

I would like to start using adeira eslint runner on a legacy project where it reports ~5k warnings. With such a high number of warnings, it's difficult to locate the error when developer breaks something. With `--no-warnings`, one can simply list only errors. Note that original eslint has `--quiet` option for this purpose, although I would prefer this name.


mrtnzlml:
Before we merge this: how are you planning on fixing these warnings? I don't think this is the best practice since it leads to simply ignoring all of the warnings. Would be better to have some other strategy, no?

tbergq:
Maybe you could simply turn the rules giving you warnings to off in your setup, then turn one and one rule on, to fix them in a sane way?

jaroslav-kubicek:
The strategy should be this:
- warnings will be still turned on by default
- when dev changes some files, he tries to resolve issues on that subset of files
- when you change some files, get errors and don't know why - you can filter them out with this option

Also, specifically speaking about Kiwi, we should have now a person who has removing warnings as one of the main objectives. 

And I would like to also experiment with one additional feature - measuring stats that no. of warnings is not increasing per rule per file.

jaroslav-kubicek:
can we merge this? 🙏

PR #294, opened 2020-01-09
eslint-config: change "no-constant-condition" from warning to error

jaroslav-kubicek:
I would like to propose making `no-constant-condition` an error. There shouldn't be an use case for the following code:

```
if (true) {
  stuff();
}
```

and constant conditions will remain allowed in loops thanks to `{ checkLoops: false }`.

This type of mistake can occur in code when a developer wants to quickly enable/disable particular branch of code, e.g.:

```
if (isTurnedOn() && false) {
  doStuff();
}
```

Such a practice should be discouraged, leading me to this suggestion.

### DIFF
--- a/src/eslint-config-adeira/ourRules.js
+++ b/src/eslint-config-adeira/ourRules.js
@@ -18,7 +18,7 @@ module.exports = {
   'no-compare-neg-zero': ERROR,
   'no-cond-assign': ERROR,
   'no-console': ERROR,
-  'no-constant-condition': [WARN, { checkLoops: false }],
+  'no-constant-condition': [NEXT_VERSION_ERROR, { checkLoops: false }],
   'no-control-regex': ERROR,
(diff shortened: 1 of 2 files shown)


src/eslint-config-adeira/ourRules.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
   'no-compare-neg-zero': ERROR,
   'no-cond-assign': ERROR,
   'no-console': ERROR,
-  'no-constant-condition': [WARN, { checkLoops: false }],
+  'no-constant-condition': [ERROR, { checkLoops: false }],
```
mrtnzlml: Can we make it a `NEXT_VERSION_ERROR` first? We usually don't jump from WARN to ERROR but give some time for the migration. 🤔

jaroslav-kubicek: ok

PR #298, opened 2020-01-11
Relay: sync 'useMutation' with Pipedrive implementation

mrtnzlml:
We not internally need to work with disposables, optimistic responses and we also synced the public interface with Relay Experimental (first the executor, second the pending status).


src/relay/src/hooks/__tests__/useMutation.test.js
```diff
@@ -0,0 +1,148 @@
+// @flow
+
+/* global document */
+
+import React, { useEffect } from 'react';
```
jaroslav-kubicek: AFAIK we always imported everything `import * as React from 'react';`

tbergq: There is nothing wrong with this import

jaroslav-kubicek: That’s true, I’m trying to point out consistency

mrtnzlml: Feel free to create an Eslint rule if you think it's essential. I think it's better to focus on other things than consistency, which in this case, doesn't bring anything.

PR #300, opened 2020-01-11
Move Relay logger into new '@adeira/relay-runtime' package

mrtnzlml:
Spring cleanup - we should split the large package into more smaller ones.


tbergq:
All for smaller packages, makes sense. But what about moving the `getDataFromRequest` from `@adeira/relay-utils` to this package. Somehow it feels like it also belongs to `runtime`.

I don't think we need multiple packages covering the same areas.

mrtnzlml:
@tbergq The purpose of this package is to expose `relay-runtime` specific things. Can we reconsider it later when the packages are clearly separated? I am still quite reluctant about these custom things since I believe they should be part of the individual project.

I'd be OK with it when they prove to be super helpful but it doesn't seem to be the case for `getDataFromRequest` or things like async commit mutation. 🤔 

Anyway, please keep pushing me if you really want it here (possibly by opening a PR with some solid arguments).

tbergq:
I thought looking up values from the store was `runtime specific`.

Also, I don't mind having this function in my own repos where I use them, I just thought the general idea was not to access `react-relay` and `relay-runtime` directly, but rather use `@adeira` packages. 

We are still early in this process, so let's keep exploring it, keeping these packages separate for now.

mrtnzlml:
> I just thought the general idea was not to access react-relay and relay-runtime directly, but rather use @adeira packages.

I agree with this. So I'd expect that this package provides `getRequest` and `createOperationDescriptor` with correct types and you can do whatever you want with it in your application. 🤔 

OK, let me think about it again. It should not be such a big deal to add it back to this new package.

PR #319, opened 2020-01-13
Relay runtime: add TS types

mrtnzlml:


src/relay-runtime/index.d.ts
```diff
@@ -0,0 +1,3 @@
+// @flow
```
jaroslav-kubicek: `@flow` in `.ts` file? 🤔

mrtnzlml: Hah, that's an interesting glitch in our Eslint config, see #320

Issue #210, opened 2019-12-03
@adeira/eslint-plugin: Introduce new rule for better GraphQL server


jaroslav-kubicek:
when implementing GraphQL server, I suggest to check that every instance of  `GraphQLObjectType` satisfies following rules:

- has a description
- name matching the filename

mrtnzlml:
One year! Happy birthday! 🎂 😊 

I would like to close this issue as it is being stale and we were not able to fix it in a reasonable time. It could be due to insufficient resources from our side or we simply do not consider it to be a big deal.

Here are some options what to do now:

1. You can send a pull request if you consider the issue to be important for you and you'd like to contribute back. ✍️ 
1. You can consider sponsoring one of our maintainers to help us deal with the associated maintenance and development costs. It will be greatly appreciated. 👍 
1. You can fork our open-source projects and start developing your own version. Consider contributing back first though.
1. Do nothing.

Thanks for understanding. ❤️

PR #336, opened 2020-01-14
eslint-plugin: Add new "graphql-require-description" rule

jaroslav-kubicek:
related to https://github.com/adeira/universe/issues/210

### DIFF
--- a/src/eslint-plugin-adeira/src/rules/__tests__/graphql-require-description.test.js
+++ b/src/eslint-plugin-adeira/src/rules/__tests__/graphql-require-description.test.js
@@ -1,0 +1,48 @@
+// @flow
+
+const RuleTester = require('eslint').RuleTester;
+
+const rule = require('../graphql-require-description');
+
+const ruleTester = new RuleTester({
+  parser: require.resolve('babel-eslint'),
+});
+
+ruleTester.run('graphql-require-description', rule, {
+  valid: [
+    `new Date`,
+    `new Pokus({ foo: { baz: 'bar' }})`,
+    `new GraphQLObjectType('Random stuff')`,
+    `
+    new GraphQLObjectType({
+      name: 'TripSegment',
+      description: '"TripSegment" represents single atomic connection between two places in itinerary.',
+      fields: {
+        departure: {
+          type: GraphQLRouteStop,
+          resolve: ({ departure }: TripData): DepartureArrival => departure,
+        },
+      },
+    });
+  `,
+  ],
+
+  invalid: [
+    {
+      code: `
+        new GraphQLObjectType({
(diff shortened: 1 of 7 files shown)


src/eslint-plugin-adeira/src/rules/__tests__/graphql-require-description.test.js
```diff
@@ -0,0 +1,48 @@
...
+          name: 'TripSegment',
+          fields: {
+            departure: {
+              type: GraphQLRouteStop,
+              resolve: ({ departure }: TripData): DepartureArrival => departure,
+            },
+          },
+        });
+  `,
+      errors: [
```
tbergq: What if I add `description: ''`

tbergq:
Not sure how I feel about this rule, since I might add empty description, or just `adfa`. 

We cannot really validate that the description is meaningful by eslint alone. And I also believe some types are self descriptive, though I agree best practise is to have a description.

mrtnzlml:
> Not sure how I feel about this rule, since I might add empty description, or just `adfa`.
> 
> We cannot really validate that the description is meaningful by eslint alone. And I also believe some types are self descriptive, though I agree best practise is to have a description.

Facebook uses something like `GraphQLSelfDescriptive` type for descriptions which are self-descriptive (which effectively makes every description required, not only for objects but even fields). See:

Screenshot 2020-01-15 at 13 21 44

tbergq:
Ok, that is also good. I am not saying I am against the rule, just that, you will still need to address this on CR. Because I might just add `desicription: 'a'` just to shut up the eslint rule.

src/eslint-plugin-adeira/src/rules/__tests__/graphql-require-description.test.js
jaroslav-kubicek: you are right, I realized it later, will fix, thanks!

jaroslav-kubicek:
yes, the description doesn't add much value for objects like `Address`, `Price` etc. and I'm fine if people just put there `User` into description for `User`. My goal is to just make people think what such object represents, improve docs and be consistent.

Graphiql even puts there `No Description` as a little complain 🙂 

Screenshot 2020-01-15 13 39 53

tbergq:
> My goal is to just make people think what such object represents, improve docs and be consistent

Yes, good things, so let's continue and merge it when the empty string is handled 😊

src/eslint-plugin-adeira/src/rules/__tests__/graphql-require-description.test.js
jaroslav-kubicek: should be fixed now!

Issue #337, opened 2020-01-15
Eslint-config: enable "no-unlimited-disable"


jaroslav-kubicek:
Currently, `eslint-comments/no-unlimited-disable` eslint rule is disabled because it would collide with files autogenerated by relay. Turn this on right when this MR is resolved: https://github.com/mysticatea/eslint-plugin-eslint-comments/pull/36

**EDIT:**

use override: https://github.com/mysticatea/eslint-plugin-eslint-comments/pull/36#issuecomment-574787370

PR #341, opened 2020-01-15
eslint-config: turn on "eslint-comments/no-unlimited-disable" rule

jaroslav-kubicek:
closes https://github.com/adeira/universe/issues/337


src/eslint-config-adeira/getCommonConfig.js
```diff
@@ -82,5 +83,15 @@ module.exports = function getCommonConfig(rules /*: EslintConfigRules */) /*: Es
       'eslint-plugin-adeira',
       'eslint-plugin-prettier',
     ],
+
+    overrides: [
+      {
+        files: ['**/__generated__/*.graphql.js'],
+        rules: {
+          // Relay disables generated files with unlimited scope
```
mrtnzlml: You know we have Relay under control and we could just generate the files with correct Eslint disables? 😅

jaroslav-kubicek: I know, I like your way of thinking, at one point I was considering it too except I don’t know if we should do that - we would disable only what doesn’t match our rules, but user might have different config. Therefore I would suggest to have this and let that idea sink in.

src/eslint-config-adeira/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Added `react/jsx-no-script-url` as next_version_error
+- `eslint-comments/no-unlimited-disable` rule is turned on, ignoring all `__generated__/*.graphql.js` files by default, so files autogenerated by Relay compiler are not reported. Follow [Eslint Docs](https://eslint.org/docs/user-guide/configuring#disabling-rules-only-for-a-group-of-files) on how to disable additional group of files if needed.
```
mrtnzlml: Can we add here the `overrides` example as well to make it easier?

jaroslav-kubicek: fixed!

PR #351, opened 2020-01-22
Relay example: fix misconfigured Shipit

mrtnzlml:
There was a logical error: you cannot try to ship some file and then strip it from the shipping process. I am not sure we actually support such file replacement. This could be (in theory) fairly simple to fix by reordering the filters.

### DIFF
--- a/src/monorepo-shipit/config/__tests__/example-relay.test.js
+++ b/src/monorepo-shipit/config/__tests__/example-relay.test.js
@@ -19,7 +19,6 @@ testExportedPaths(path.join(__dirname, '..', 'example-relay.js'), [
 
   // invalid cases:
   ['src/example-relay/.babelrc.js', undefined], // correctly deleted
-  ['src/example-relay/jest.config.js', undefined], // correctly deleted
   ['src/example-relay/__github__/unknown.js', undefined], // correctly deleted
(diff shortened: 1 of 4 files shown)


src/monorepo-shipit/config/__tests__/example-relay.test.js
```diff
@@ -19,7 +19,6 @@ testExportedPaths(path.join(__dirname, '..', 'example-relay.js'), [
 
   // invalid cases:
   ['src/example-relay/.babelrc.js', undefined], // correctly deleted
-  ['src/example-relay/jest.config.js', undefined], // correctly deleted
```
mrtnzlml: @jaroslav-kubicek Guess who wrote this test ensuring `jest.config.js` is never going to be published via Shipit? 😅

mrtnzlml:
> This could be (in theory) fairly simple to fix by reordering the filters.

It's actually not that simple. We have to first move the directories and then strip the files (otherwise we would strip them before having chance to move them). So I am not sure how to technically achieve file replacement with the current tools. We'd probably need to add support for it.

src/monorepo-shipit/config/__tests__/example-relay.test.js
jaroslav-kubicek: nooo :)

PR #353, opened 2020-01-23
eslint: Less strict warning log for dependency

tbergq:
We should only log when version is less than expected
or if it is a major version greater than expected.
Using 7.18.0 when expected is 7.17.0 should not be a problem.


mrtnzlml:
> Using 7.18.0 when expected is 7.17.0 should not be a problem.

I think this can actually cause a problem. It's because you could have 2 versions of the plugin and Eslint sometimes does some weird errors because of these 2 instances. That was the initial idea for this warning.

tbergq:
> It's because you could have 2 versions of the plugin

I think you can have two versions also with the `!==` check, we are not checking against yarn.lock here, but package.json. We can never be sure what is actually in the users project.

tbergq:
So, what if we also run a check to see if there are different versions installed?

mrtnzlml:
> So, what if we also run a check to see if there are different versions installed?

Do you know how to do it better? This check is literally only for this - nothing else.

tbergq:
> Do you know how to do it better?

Nothing comes to mind, I will try to thing about it. Closing this PR for now. Thanks for explaining.

PR #360, opened 2020-01-26
eslint: Add note about installing jest in readme

### LINKED ISSUE #357: Eslint runner fails without jest installed
`error Command "jest" not found.`

tbergq:
You cannot use the eslint-runner without jest installed.
We could add it to package.json, but this might conflict
with the version of jest the consumer already uses.

closes #357


src/eslint-config-adeira/README.md
```diff
@@ -66,6 +66,12 @@ module.exports = {
 };
 ```
 
+Since the eslint-runner uses jest, you have to install it if you did not already do it:
+
+```
+yarn add jest --dev [--ignore-workspace-root-check]
```
mrtnzlml: This should be added to the `"Install this package"` section just like we added there the Eslint. Jest dependency should be defined in peer deps. Could you please update it?

tbergq: Ok, but you can use the preset without jest, you cannot use the eslint-runner without jest. That is why I thought it belonged here.

tbergq: 🚀

mrtnzlml: OK, that's actually a good point. Can you revert it then? Sorry. 😅 

I was thinking that maybe we should separate the runner from the config completely.

tbergq: 🤦‍♂

tbergq: ✅ Thankfully, it is hard to get rid of things with git 😄

tbergq: > I was thinking that maybe we should separate the runner from the config completely.

It could be a part of monorepo-utils?

mrtnzlml: > It could be a part of monorepo-utils?

It could be a new NPM package. The idea is to be able to use it even though you are not using our config. It's not necessarily related to it.

tbergq: Yeah, that is a good point 👍

tbergq: Yeah, that is a good point 👍

tbergq: So, is it ready to be merged?

PR #361, opened 2020-01-26
eslint: Lint all if getChangesToTest fails

### LINKED ISSUE #358: Eslint runner failes if not used in a git repo
`Command failed with exit code 128: fatal: not a git repository (or any of the parent directories): .git`

### LINKED ISSUE #359: Eslint runner fails if there is no remote
` Command failed with exit code 128: fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.`

tbergq:
If git is not initialised or if there was no remote git repo
this would crash and eslint-runner did not work.
Let's rather fallback to lint all files
closes #358
closes #359


src/eslint-config-adeira/runner/run.js
```diff
@@ -33,16 +33,28 @@ type Options = {|
...
 //  implementation with this built-in mechanism. This way we don't have to iterate all the files
 //  and skipping them.
-const changedFiles = Git.getChangesToTest();
+const getChangedFiles = () => {
+  try {
+    // This fails if git is not initialised or if user has no remote origin
+    // Let's catch the error, and just lint all instead rather than crashing
+    return Git.getChangesToTest();
+  } catch {
+    return null;
```
mrtnzlml: We could also return an empty array `[]` to simplify the conditions bellow. It's strange that `changedFiles` can be an array but sometimes also a `null`. I'd prefer to annotate this `getChangedFiles` function to always return `$ReadOnlyArray<string>`

tbergq: I thought about this, it would lead to `let runAll = extraOptions.runAll || changedFiles.length === 0;`

Meaning you would lint all if you didn't change any files

PR #362, opened 2020-01-26
Cleanup exact flow types

### LINKED ISSUE #90: Cleanup exact Flow types
This essentially means to get rid of `{| |}` syntax. Preferably via Eslint rule, so it's enforced and auto-fixable.

tbergq:
closes #90


mrtnzlml:
@tbergq Out of curiosity: did you use any tool for this or simple find+replace?

tbergq:
> @tbergq Out of curiosity: did you use any tool for this or simple find+replace?

Just VS-code search and replace

.flowconfig
```diff
@@ -29,7 +29,6 @@
 ; https://flow.org/en/docs/config/lints/
 [lints]
 all=error
-ambiguous-object-type=off
```
mrtnzlml: Is this ON by default now?

src/example-relay/src/Homepage/__generated__/HomepageQuery.graphql.js
```diff
@@ -10,16 +10,16 @@ import type { ConcreteRequest } from 'relay-runtime';
 type LocationsPaginatedBidirectional_data$ref = any;
 type LocationsPaginatedRefetch_data$ref = any;
 type LocationsPaginated_data$ref = any;
-export type HomepageQueryVariables = {|
```
mrtnzlml: 🤔 We should not change this file manually. Or any other generated file.

.flowconfig
tbergq: I am a little bit confused here, I had to remove it from strict, but I can see it gives no errors when it is off, and no errrors when on. 

On the strict config I got errors 🤔

tbergq: It says `Like implicit-inexact-object, except triggers even when the exact_by_default option is set to true.`

Which seems like it does not, so, I can leave it to off

src/example-relay/src/Homepage/__generated__/HomepageQuery.graphql.js
tbergq: Good point, I should have left these filtered out.

tbergq: Should be back to normal now

mrtnzlml: I can still see some __generated__ files modified in this PR. For example: `src/example-relay/src/LocalForm/__generated__/LocalFormQuery.graphql.js`

tbergq: Sorry, I thought just running the relay-compiler was enough. They should be gone now.

mrtnzlml:
@tbergq Do we have some Eslint rule which would prevent us from using this syntax?

tbergq:
No, I don’t think there is, using the | is still valid syntax, just not necessary

michalsanger:
The rule would be handy. You could use the `|` syntax only when you explicitly ignore that rule.

tbergq:
Not really sure, I am even starting to doubt that we should merge this one. 

Since `{||}` always means exact, by removing it, we are also assuming that our consumers are using exact by default, which they are maybe not doing.

While keeping the | syntax, it will be exact no matter what flowconfig the consumer is using

mrtnzlml:
@tbergq You mean it would not be compatible with the other repositories which are not yet switched? Since we are providing the NPM libraries? So we should stick with the `{| |}` syntax for as long as possible until it's a real problem?

That sounds like a very valid and smart argument why to close this PR, yes. We are currently enforcing the `{| |}` syntax, correct?

tbergq:
> You mean it would not be compatible with the other repositories which are not yet switched?

Exactly yes. 

> We are currently enforcing the {| |} syntax, correct?

Yes, we are, through the require explicit inexact eslint rule.

So I am closing this PR 😊

PR #384, opened 2020-02-03
Release @adeira/relay-utils 0.3.0

tbergq:


tbergq:
I got flow-error in my monorepo when upgrading `@adeira/relay`, so I need a version of `@adeira/relay-utils` using the same version of `@adeira/relay`

PR #400, opened 2020-02-08
Use next api to set up graphql server

tbergq:
This should already be enough to run all location examples.
We need to also add the currency to cover the polling example.
When all is developed, we can just flip the fetch function
to target this graphql server

### DIFF
--- a/src/example-relay/src/graphql/Schema.js
+++ b/src/example-relay/src/graphql/Schema.js
@@ -1,0 +1,18 @@
+// @flow
+
+import { GraphQLSchema, GraphQLObjectType } from 'graphql';
+
+import locations from './queries/Locations';
+
+const RootQuery = new GraphQLObjectType({
+  name: 'RootQuery',
+  description: 'Root Query',
+  fields: {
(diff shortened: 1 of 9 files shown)


src/example-relay/src/graphql/Schema.js
```diff
@@ -0,0 +1,18 @@
...
+
+import { GraphQLSchema, GraphQLObjectType } from 'graphql';
+
+import AllLocations from './queries/AllLocations';
+
+const RootQuery = new GraphQLObjectType({
+  name: 'RootQuery',
+  description: 'Root Query',
+  fields: {
+    allLocations: AllLocations,
```
michalsanger: Do we need to keep this? Today we would name the query just `locations` I guess.

tbergq: Good idea, yeah, I was thinking to match the names of our current schema, but we can rather update the client when we do the switch 👍

tbergq: 🚀

PR #402, opened 2020-02-09
npm-publisher: Use projects babel config

### LINKED ISSUE #196: monorepo-npm-publisher: support custom babel.config.js
I would like to use `@adeira/monorepo-npm-publisher` on different project, but it needs to be transpiled with additional babel plugins

Proposed solution:
- publisher will try to find `babel.config.js`
- if found, it will automatically add `'@adeira/babel-preset-adeira` if it's not present in presets
- if config already contains `'@adeira/babel-preset-adeira` preset, it will be replaced as follow to enable multiple targets: 
```
[
          {
            target,
          },
 ],
```

tbergq:
Use rootMode upwards to apply the babel config from the project
using this package, rather than force the @adeira/babel-preset

closes #196

### DIFF
--- a/src/monorepo-npm-publisher/src/transformFileVariants.js
+++ b/src/monorepo-npm-publisher/src/transformFileVariants.js
@@ -11,9 +11,8 @@ export default function transformFileVariants(
   transpileESM: boolean,
 ): void {
   const getBabelConfig = (target: 'js' | 'js-esm' | 'flow') => {
-    // TODO: load Babel config from the packages instead (if exists)
     return {
-      root: __dirname, // do not lookup any other Babel config
+      rootMode: 'upward',
       presets: [
         [
           '@adeira/babel-preset-adeira',
--- a/src/monorepo-npm-publisher/CHANGELOG.md
+++ b/src/monorepo-npm-publisher/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Set root mode to upwards to apply consumer projects babel config
+
 ---
 
 Changelog before our fork:


tbergq:
I tested it by creating a new monorepo, where I added `@kiwicom/orbit-components` with `@kiwicom/babel-plugin-orbit-components` and the transpiled version build `.mjs`, `.js` and `.js.flow` and also applied the changes from `@kiwicom/babel-plugin-orbit-components`. `import { Button } from '@kiwicom/orbit-components'` was transpiled to `var _Button = _interopRequireDefault(require("@kiwicom/orbit-components/lib/Button"));`

src/monorepo-npm-publisher/src/transformFileVariants.js
```diff
@@ -11,9 +12,9 @@ export default function transformFileVariants(
   transpileESM: boolean,
 ): void {
   const getBabelConfig = (target: 'js' | 'js-esm' | 'flow') => {
-    // TODO: load Babel config from the packages instead (if exists)
     return {
-      root: __dirname, // do not lookup any other Babel config
+      root: findMonorepoRoot(),
+      rootMode: 'upward',
```
jaroslav-kubicek: if you set root as monorepo root, isn't `rootMode=upward` same as `rootMode=root`?

tbergq: Yes, I think you are right, I will remove the `root`

jaroslav-kubicek:
I'm just curious about last thing - how it works when adeira preset is already in the config?

tbergq:
I ran it in our monorepo with dryRun: true. And as far as I could tell, it still worked as expected.

PR #404, opened 2020-02-10
Add currency query with just random rate

tbergq:


src/example-relay/src/graphql/types/CurrencyDetail.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+
+import { GraphQLObjectType, GraphQLString, GraphQLFloat } from 'graphql';
+import globalIdField from '@adeira/graphql-global-id';
+
+export default new GraphQLObjectType({
+  name: 'CurrencyDetail',
```
michalsanger: Just a `Currency` maybe?

tbergq: 🚀

mrtnzlml:
🚀

PR #407, opened 2020-02-10
Use own implementation of connectionFromArray

### LINKED ISSUE #406: Example: bi-directional pagination doesn't work correctly
See: https://relay-example.adeira.now.sh/

It's possible to go to the next page but not to the previous page. It's just a small bug related to our custom server migration.

tbergq:
Current implementation in graphql-js is broken.
This the the solution we used in graphql.kiwi.com to patch it
closes #406

### DIFF
--- a/src/example-relay/src/graphql/queries/Locations.js
+++ b/src/example-relay/src/graphql/queries/Locations.js
@@ -1,24 +1,50 @@
 // @flow
 
-import { connectionArgs, connectionFromArray } from 'graphql-relay';
+import {
+  connectionArgs,
+  cursorToOffset,
+  connectionFromArray as relayConnectionFromArray,
+  offsetToCursor,
+  type Connection,
+  type ConnectionArguments,
+} from 'graphql-relay';
 
 import locations from '../datasets/locations.json';
 import LocationConnection from '../types/LocationConnection';
 
-type Args = {
-  +first?: number,
-  +last?: number,
-  +after?: string,
-  +number?: string,
-  ...
-};
+// There is open discusion how to handle bidirectional pagination:
+// https://github.com/graphql/graphql-relay-js/issues/58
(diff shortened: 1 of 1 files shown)


michalsanger:
I remember the code :-) Why not put it in `ArrayConnection.js` and include also tests?

tbergq:
Yeah, I want to do some more cleanup, maybe put graphql into its own workspace, start a `graphql-utils` workspace where this array connection could live. I want to look into that in the weekend. For now, I just wanted to quickly fix the pagination that broke when switching server.

mrtnzlml:
> start a graphql-utils workspace where this array connection could live

Didn't we have this package? Did we remove it since it was not needed? 😬

tbergq:
> Didn't we have this package? Did we remove it since it was not needed?

Yes, I think maybe we did. We should be able to retrieve it from the gitlog then

mrtnzlml:
Yeah, I still have the full backup so I can restore it. Create an issue please if needed.

Also, I'd choose to use `graphql-relay` or our utils but not both. It's weird to remember how to mix it correctly in order to make it work. The GraphQL Relay package has other issues in my opinion we could fix. For example: it works with `string` cursors - that could be fixed.

tbergq:
> Also, I'd choose to use graphql-relay or our utils but not both

I agree, also they are not using explicit inexact objects in flow 🙈 causing some issues. I was thinking maybe sending a PR to that repo to fix it. I would go with our utils

>  I still have the full backup so I can restore it

It should be possible to restore it from the git log, but I will ping you if. I cannot find it.

mrtnzlml:
OK. I wouldn't expect you to find it there. I cleaned these parts from history because some code was Kiwi specific. Unless we deleted it from Adeira, of course.

Issue #421, opened 2020-02-14
Upgrade to Relay 9.0.0


mrtnzlml:
Don't forget to run the Relay Compiler! :)

PR #422, opened 2020-02-14
Upgrade relay

tbergq:
Closes: https://github.com/adeira/universe/issues/421

### DIFF
--- a/src/relay/src/hooks/useRelayEnvironment.js
+++ b/src/relay/src/hooks/useRelayEnvironment.js
@@ -1,9 +1,20 @@
 // @flow
 
-import { useRelayEnvironment as _useRelayEnvironment } from 'react-relay/hooks';
+import * as React from 'react';
+import { ReactRelayContext } from 'react-relay';
+import { invariant } from '@adeira/js';
 
 import type { Environment } from '../runtimeTypes.flow';
 
 export default function useRelayEnvironment(): Environment {
-  return _useRelayEnvironment();
+  const context = React.useContext(ReactRelayContext);
+  invariant(
+    context != null,
+    'useRelayEnvironment: Expected to have found a Relay environment provided by' +
+      'a `RelayEnvironmentProvider` component or `QueryRenderer`. This usually means that' +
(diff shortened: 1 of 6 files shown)


src/relay/src/hooks/useRelayEnvironment.js
```diff
@@ -1,9 +1,16 @@
...
 import type { Environment } from '../runtimeTypes.flow';
 
 export default function useRelayEnvironment(): Environment {
-  return _useRelayEnvironment();
+  const context = React.useContext(ReactRelayContext);
+  invariant(
+    context != null,
+    'Expected to be used inside a relay context, but was not. Make sure it is wrapped in a RelayEnvironmentProvider or QueryRenderer',
+  );
+  return context.environment;
```
mrtnzlml: This is what we used before (basically just a better error message):

```js
// @flow

import { useContext } from 'react';
import { ReactRelayContext } from 'react-relay';
import { invariant } from '@adeira/js';

import type { Environment } from '../runtimeTypes.flow';

/**
 * You can use this hook in any component below QueryRenderer or RelayEnvironmentProvider.
 */
export default function useRelayEnvironment(): Environment {
  const context = useContext(ReactRelayContext);
  invariant(
    context != null,
    'useRelayEnvironment: Expected to have found a Relay environment provided by' +
      'a `RelayEnvironmentProvider` component or `QueryRenderer`. This usually means that' +
      'useRelayEnvironment was used in a component that is not a descendant of these components. ' +
      'Please make sure a `RelayEnvironmentProvider` has been rendered somewhere as a parent of ' +
      'your component. It should be defined only once in the whole application.',
  );
  return context.environment;
}
```

tbergq: Sure, I will use the one from before

tbergq:
> No changes in generated files? Did you try to run Relay Compiler on our example? :)

No changes, I ran it both in example and in relay

src/relay/src/hooks/useRelayEnvironment.js
tbergq: @mrtnzlml please recheck with new error message 🚀

PR #426, opened 2020-02-16
Move graphql code to separate workspace

tbergq:


tbergq:
Yeah, I was even considering example-relay-graphql. 

Maybe we should have a better name for it before merging.

mrtnzlml:
👍 for `example-relay-graphql` - it's a best practice to use a common prefix for related projects in our setup.

src/graphql/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@adeira/graphql",
+  "private": false,
```
mrtnzlml: Should be private, no? It's not being published to NPM.

src/graphql/.npmignore
```diff
@@ -0,0 +1 @@
+__tests__
```
mrtnzlml: This should not be needed for private packages.

tbergq:
If it is not on npm, how will it be deployed as a part of the relay-example project?

tbergq:
Closing, we will keep it within the `example-relay` for now

Issue #436, opened 2020-02-18
monorepo-utils: "getTouchedWorkspaces" fails when run without yarn


jaroslav-kubicek:
Consider a script in file `example.js` with the following code:

```
const { getTouchedWorkspaces } = require('@adeira/monorepo-utils');

const touchedWorkspaces = getTouchedWorkspaces();
console.log(touchedWorkspaces);
```

Running such file directly via node, `node example.js`, fails on unexpected char when we are parsing stdout of executed shell command as JSON.

The problem is that execution of `new ShellCommand(null, 'yarn', 'workspaces', 'info', '--json')`  -> execution of yarn on "top level" outputs additional noise:

```
yarn workspaces v1.22.0
{}
✨  Done in 0.05s.
```

mrtnzlml:
What kind of failure is it? I don't understand it from the description.

@tbergq Didn't you fix something related to this issue?

tbergq:
Not quite sure I understand it either. 

You can see what I did in [this commit](https://github.com/adeira/universe/commit/181b6a47f00e72e2f366b1fbbc99189b593f81b9). Not sure it was related, but it was somehow necessary since the return of `yarn workspaces info --json` changed with yarn 1.22.0.

mrtnzlml:
@jaroslav-kubicek Are you using `yarn policies set-version` and enforcing the right Yarn version?

tbergq:
Ok, I understand the issue. Given that you write code that doesn't need to be transpiled, 

```
// test.js
const { getTouchedWorkspaces } = require('@adeira/monorepo-utils');

const touchedWorkspaces = getTouchedWorkspaces();
console.log(touchedWorkspaces);
```
and you run it with node test.js it fails

tbergq:
Then running it with `yarn monorepo-babel-node test.js` works.

tbergq:
I think this is related to the transpiled version of  ShellCommand rather than `getTouchedWorkspaces`

jaroslav-kubicek:
oh damn, I hope I'm at least by 1 % better dev than the writer 🤦‍♂

PR #441, opened 2020-02-19
monorepo-utils: sanitize shell output of "yarn workspaces info"

jaroslav-kubicek:
fixes https://github.com/adeira/universe/issues/436

### DIFF
--- a/src/monorepo-utils/src/getWorkspaceDependencies.js
+++ b/src/monorepo-utils/src/getWorkspaceDependencies.js
@@ -5,17 +5,17 @@ import ShellCommand from './ShellCommand';
 import type { WorkspaceDependencies } from './Workspaces.flow';
 
 /**
- * This function returns the workspace dependenciesl
+ * This function returns the workspace dependencies
  */
 export default function getWorkspaceDependencies(): WorkspaceDependencies {
-  const stdout = new ShellCommand(null, 'yarn', 'workspaces', 'info', '--json')
+  const stdout = new ShellCommand(null, 'yarn', '--json', 'workspaces', 'info')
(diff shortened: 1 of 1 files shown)


tbergq:
Though this probably fixes this issue with `yarn workspaces info --json`, this could happen with other scripts using `ShellCommand`and getting stdout, so it would be interesting to find out why it is including the rest of the output when it is run with `node` rather than with `yarn mononrepo-babel-node`

jaroslav-kubicek:
AFAIK it's how yarn behaves, when you run `yarn workspaces info --json` you get exactly the same, quite unfortunate since you are asking for json, so it's not `ShellCommand` problem, only this particular case.

tbergq:
But somehow, when you create a file with `ShellCommand(null, 'yarn', 'workspaces, 'info', '--json')`

And run that file with `yarn monorepo-babel-node file.js`, I would like to know why that works, and `node file.js` does not.

jaroslav-kubicek:
alright, I think I got that 😅

mrtnzlml:
> alright, I think I got that 😅

Do you know what is the problem? Sounds like a bug in Yarn since `yarn workspaces info --json` and `yarn workspaces info` returns the same output. 🤔

tbergq:
> > alright, I think I got that 😅
> 
> Do you know what is the problem? Sounds like a bug in Yarn since `yarn workspaces info --json` and `yarn workspaces info` returns the same output. 🤔

It seems the correct usage is `yarn --json workspaces info`

mrtnzlml:
@tbergq That sounds like a solution to our problem.

PR #454, opened 2020-02-24
relay-example: Add range add mutation example

tbergq:


tbergq:
> Just a suggestion to add few words on top of the example page explaining what's this about.

Good idea, I will add it before we merge

tbergq:
@michalsanger, looks ok?

michalsanger:
@tbergq just not sure about the `Alert`. I used this markup:
https://github.com/adeira/universe/blob/master/src/example-relay/src/LocalForm/index.js#L78-L85

tbergq:
Ok, I will use the same

tbergq:
better @michalsanger ?

michalsanger:
@tbergq the best :-)

src/example-relay/src/mutations/RangeAdd/LocationsForm.js
```diff
@@ -0,0 +1,111 @@
...
+  const [state, dispatch] = React.useReducer(reducer, initialState);
+  const [loading, setLoading] = React.useState(false);
+  const env = useRelayEnvironment();
+  const onSubmit = (e: SyntheticEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    if (state.type === '') {
+      alert('You must select a type');
+      return;
+    }
+    setLoading(true);
```
mrtnzlml: You should use our `useMutation` hook instead. It returns `isPending` boolean so you don't have to deal with it manually like this.

PR #457, opened 2020-02-25
Expose use mutation from @adeira/relay root

tbergq:
Release new version of packages depending on this change.

### DIFF
--- a/src/relay/CHANGELOG.md
+++ b/src/relay/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+# 1.1.0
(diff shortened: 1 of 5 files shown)


src/relay/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+# 1.0.1
```
mrtnzlml: This should be a minor I think.

tbergq: Ok, I will fix

PR #472, opened 2020-02-28
monorepo-utils: Correctly log timezone

tbergq:
Currenlty it was logging UTC, even if you had another
timezone in process.env.TZ

### DIFF
--- a/src/monorepo-utils/src/TestsRunner.js
+++ b/src/monorepo-utils/src/TestsRunner.js
@@ -26,7 +26,7 @@ function _runJest(config, timezone = 'UTC') {
     }
   });
 
-  console.warn(`Running tests in timezone: ${timezone}`); // eslint-disable-line no-console
+  console.warn(`Running tests in timezone: ${process.env.TZ ?? timezone}`); // eslint-disable-line no-console
   return new ShellCommand(
     null,
     'node',


mrtnzlml:
These timezones work a bit differently. There are 3 timezones available and cannot really be changed:

```js
const timezones = [
  'UTC',
  'Asia/Tokyo', // +9
  'America/Lima', // -5
];
```

The idea was, that you run the tests with `parallel:3` in GitLab and it automatically runs them in these timezones to verify everything works as expected (see `process.env.CI_NODE_TOTAL` and `process.env.CI_NODE_INDEX`).

tbergq:
Yes, well, I tried to run them locally in another project with `TZ=America/Lima yarn test` and it kept logging `Running tests in timezone UTC` which was very confusing

tbergq:
Because the test was actually run in the correct timezone, just the log was wrong.

mrtnzlml:
> Because the test was actually run in the correct timezone, just the log was wrong.

I believe this is an unintentional bug that turned into a feature... 😅

tbergq:
> > Because the test was actually run in the correct timezone, just the log was wrong.
> 
> I believe this is an unintentional bug that turned into a feature... 😅

Intentionally or not, It is nice to be able to run tests in the timezone of your choice locally

PR #500, opened 2020-03-16
monorepo-utils: Add "getChangedFilesByFilter" to filter changed files

jaroslav-kubicek:
I welcome any suggestion on how to make API in the best possible way 🙂 

This function executes `git diff --name-only --diff-filter=filters origin/master...HEAD`, I thought that individual letters which you can pass as a filter (e.g. 'AM' for added and modified files) could be passed in an array (`['A', 'M']`) instead of a single string so we can type it.

### DIFF
--- a/src/monorepo-utils/src/Git.js
+++ b/src/monorepo-utils/src/Git.js
@@ -4,6 +4,30 @@ import os from 'os';
 
 import ShellCommand from './ShellCommand';
 
+/*::
+
+type DiffFilter =
+  | 'A' // added files
+  | 'C' // copied
+  | 'D' // deleted
+  | 'M' // modified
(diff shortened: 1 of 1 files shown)


jaroslav-kubicek:
hmm hmm, ok, I will think about it, would be nice to keep API as simple as possible, `--cached` can't be used when comparing changes in the branch since the common ancestor

mrtnzlml:
@jaroslav-kubicek What about the PR? Should we merge it? Should we close it?

jaroslav-kubicek:
I suggest to close it - I decided I would prefer to keep API as simple as possible - instead of this, I will probably extend already existing functions to accepts optional args, WDYT?

PR #527, opened 2020-03-21
Babel preset: use named snapshots

mrtnzlml:


mrtnzlml:
cc @tbergq - should be OK for review

PR #533, opened 2020-03-23
Move Record proxy type from relay to relay-runtime

tbergq:
Using connection handler gave flow error because of mismatch in Record proxy type

### DIFF
--- a/src/relay-runtime/src/RelayRuntimeTypes.js
+++ b/src/relay-runtime/src/RelayRuntimeTypes.js
@@ -6,19 +6,22 @@ export type Variables = { +[name: string]: $FlowFixMe, ... };
 
 export type DataID = string;
 
-export interface RecordProxy {
-  copyFieldsFrom(source: RecordProxy): void;
-  getDataID(): DataID;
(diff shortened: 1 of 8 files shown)


src/relay-runtime/src/RelayRuntimeTypes.js
```diff
@@ -6,19 +6,23 @@ export type Variables = { +[name: string]: $FlowFixMe, ... };
 
 export type DataID = string;
 
-export interface RecordProxy {
-  copyFieldsFrom(source: RecordProxy): void;
-  getDataID(): DataID;
-  getLinkedRecord(name: string, args?: ?Variables): ?RecordProxy;
```
mrtnzlml: @tbergq Why cannot we use the `Variables` type here? It's already in Runtime package.

tbergq: Probably not, this is just the result of a hasty copy-paste. I will fix it.

tbergq: 🚀

PR #559, opened 2020-03-31
GraphQL ID: rename `__isTypeOf` to `isTypeOf`

mrtnzlml:
Preparation for stable release.

### DIFF
--- a/src/graphql-global-id/README.md
+++ b/src/graphql-global-id/README.md
@@ -50,7 +50,11 @@ const fields = {
 };
 ```
 
-# Restoring original ID
+# `toGlobalId`
+
+TKTK
+
+# Restoring original ID (`fromGlobalId``)
 
 `GlobalID` used in previous examples accepts readable (unmasked ID) and returns opaque ID. You can revert this process in case you need to access the original unmasked ID. Check the following test to get the idea:
 
@@ -63,6 +67,10 @@ it('returns correct original ID', () => {
 ```
 
 You should always use this function because the way how this utility works with masking and unmasking is just an internal detail.
+
+# `isTypeOf`
+
+TKTK
 
 # Tips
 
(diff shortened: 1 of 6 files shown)


mrtnzlml:
Should be without breaking changes so people can migrate the code.

src/graphql-global-id/src/GlobalID.js
```diff
@@ -27,8 +27,15 @@ export function toGlobalId(type: string, id: string | number): OpaqueIDString {
   return encode(`${type}:${id}`);
 }
 
-// TODO: find out better way how to do it (type should be just an internal detail - see evaluateGlobalIdField)
+/**
+ * TODO: remove in 1.0.0
+ * @deprecated
+ */
 export function __isTypeOf(type: string, opaqueID: mixed): boolean {
+  return isTypeOf(type, opaqueID);
```
tbergq: How about also using `warning` and log here that it is deprecated?

mrtnzlml: Do you think it's necessary? I think we have the usages somehow under control, or not? :) cc @michalsanger

tbergq: necessary, probably not, but I think it is good practise to log this in dev mode. To remind any consumers that it has been deprecated. 

Ok, we don't have that many outside of the 4 of us. So I am also fine with merging it as is.

src/graphql-global-id/README.md
```diff
@@ -50,7 +50,11 @@ const fields = {
 };
 ```
 
-# Restoring original ID
+# `toGlobalId`
+
+TKTK
```
jaroslav-kubicek: what's this? 😄 

Screenshot 2020-03-31 17 44 47

mrtnzlml: - https://en.wikipedia.org/wiki/To_come_(publishing)
- https://www.urbandictionary.com/define.php?term=TKTK

PR #569, opened 2020-04-03
Replace @kiwicom/orbit-components with grommet

tbergq:


src/example-relay/src/Homepage/locations/Location.js
```diff
@@ -20,12 +18,11 @@ function Location({ location }: Props) {
   const name = location.name ?? '';
   return (
     <li>
-      <Stack direction="row" spacing="compact">
-        <CountryFlag location={location} />
```
mrtnzlml, line 24: I believe there was something like `iconUrl` which you could use here. But no big deal.

src/example-relay/src/Homepage/locations/LocationsPaginated.js
```diff
@@ -32,9 +32,7 @@ function LocationsPaginated(props: Props) {
           <Location key={edge?.node?.id} location={edge?.node} />
         ))}
       </ol>
-      <Button onClick={loadMore} size="small">
-        Load more!
-      </Button>
+      <Button label=" Load more!" onClick={loadMore} size="small" primary />
```
mrtnzlml, line 35:
```suggestion
      <Button label="Load more!" onClick={loadMore} size="small" primary />
```

src/example-relay/src/Homepage/locations/Location.js
tbergq: Ahh, yes, we have a `countryFlagURL`, I see if I can re add it in a later PR

tbergq:
> Nice one! 👍
> 
> How was your experience with `grommet`?

Yeah, it was good, I would definitely use it again 👍 

The most difficult was to find out how to extend the button styles 😂

PR #573, opened 2020-04-06
Relay-example: use js-esm as build target

### LINKED ISSUE #572: Fix deployment of example relay
After replacing orbit-components with grommet, we get a runtime exception

Screenshot 2020-04-06 at 08 52 18

tbergq:
closes #572

### DIFF
--- a/src/example-relay/next.config.js
+++ b/src/example-relay/next.config.js
@@ -11,5 +11,13 @@ module.exports = withCustomBabelConfigFile(
       // @x-shipit-enable: 'babel.config.js',
       '.babelrc', // @x-shipit-disable
     ),
+    webpack: config => {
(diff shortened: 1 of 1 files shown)


tbergq:
> Do you know why it didn't work before? 🤔

Not 100% sure, but from the error message, it seems like it expects to load commonjs, while actually loading `es-module`, and that causes the error.

tbergq:
> Feels quite hacky while seems like we don't understand very much what's going on. 😕

Yeah, still, only way I am able to make it work so 🤷‍♂ I really don't know what is going on behind the scenes with babel, maybe if I did, I would find a better solution.

tbergq:
Hmm, ok, I might have found a better solution. ⏳

PR #585, opened 2020-04-08
Improved evaluateGraphQLResolver

michalsanger:
Fields without explicit resolver are supported now.
It also calls `serialize` method if the field has one (scalars does) so invalid inputs can be tested.

### DIFF
--- a/src/test-utils/src/evaluateGraphQLResolver.js
+++ b/src/test-utils/src/evaluateGraphQLResolver.js
@@ -1,0 +1,34 @@
+// @flow
+
+/**
+ * Use this function to evaluate resolvers in test files. Usage:
+ *
+ * ```js
+ * const fields = Location.getFields();
+ * expect(
+ *   evaluateGraphQLResolver(fields.countryFlagURL, {
+ *     country: ' ... ', // test value
+ *   }),
+ * ).toBe(' ... ');
+ * ```
+ */
+export default function evaluateGraphQLResolver(
+  field: { [key: string]: any, name: string, ... },
+  testValue: { [key: string]: any, ... },
+  argsValue?: { [key: string]: any, ... },
+  contextValue?: { [key: string]: any, ... },
+): any {
+  const resolveFn =
+    field.resolve ||
+    function resolveMock(): string {
+      return testValue[field.name];
+    };
+  const resolvedValue = resolveFn(testValue, argsValue, contextValue, {
+    path: { key: 'mocked' },
+  });
+
+  if (typeof field.type.serialize === 'function') {
+    return field.type.serialize(resolvedValue);
+  }
+  return resolvedValue;
+}
--- a/src/test-utils/package.json
+++ b/src/test-utils/package.json
@@ -4,11 +4,14 @@
   "homepage": "https://github.com/adeira/universe/tree/master/src/test-utils",
   "license": "MIT",
   "private": false,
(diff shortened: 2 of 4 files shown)


src/test-utils/src/evaluateGraphQLResolver.js
```diff
@@ -0,0 +1,34 @@
...
+export default function evaluateGraphQLResolver(
+  field: { [key: string]: any, name: string, ... },
+  testValue: { [key: string]: any, ... },
+  argsValue?: { [key: string]: any, ... },
+  contextValue?: { [key: string]: any, ... },
+): any {
+  const resolveFn =
+    field.resolve ||
+    function resolveMock(): string {
+      return testValue[field.name];
```
tbergq, line 24: Why do we need this fallback value, shouldn't all fields have a resolve function?

tbergq: 🤔  Oh, I guess it can be defined without one, and this is the default function 👍 

rubber-🦆-ing out loud here 😂

mrtnzlml: But does this make sense then? Why would you call `evaluateGraphQLResolver` for no resolver?

tbergq: I mean, you define a field like this: 

```
format: {
      type: GraphQLString,
      description: 'Format string of the currency',
    },
```

id doesn't have a resolver, but by default it returns parent.format

michalsanger: Because it's confusing. In the test you get `undefined` but in real usage you get some value.

tbergq: It is not only confusing, it is wrong if it works differently in the test than in the real usage

mrtnzlml: > it doesn't have a resolver, but by default it returns parent.format

@tbergq That's however changeable behavior, see `fieldResolver` in the Apollo server for example. I am not convinced it's a good strategy to simply hardcode it like this.

michalsanger: You mean resolvers map in Apollo server?

tbergq: Screenshot 2020-04-08 at 16 05 38

@mrtnzlml you mean ☝️

mrtnzlml: No, I mean `fieldResolver`: https://github.com/apollographql/apollo-server/blob/377be257b31d85a14dd2b001496194ed04ac2e03/packages/apollo-server-core/src/graphqlOptions.ts#L33

The same field exists in GraphQL.js for example. It would be at least nice to reuse the same logic if we already decided differently. Here it is (can be imported): https://github.com/graphql/graphql-js/blob/278bde0a5cd71008452b555065f19dcd1160270a/src/execution/execute.js#L1181-L1199

tbergq: Aha, nice, Yeah, I agree, lets use that one instead

michalsanger: OK, let's move here please https://github.com/adeira/universe/pull/588

PR #588, opened 2020-04-08
evaluateGraphQLResolver support non null fields

michalsanger:


src/test-utils/package.json
```diff
@@ -4,14 +4,12 @@
...
+  "version": "0.4.1",
   "main": "src/index",
   "sideEffects": false,
   "dependencies": {
     "@adeira/js": "^1.2.0",
-    "@babel/runtime": "^7.9.2"
-  },
-  "devDependencies": {
+    "@babel/runtime": "^7.9.2",
     "graphql": "^14.6.0"
```
mrtnzlml, line 13: Why is this necessary? 🤔 Shouldn't it be a `peerDependency`?

src/test-utils/src/evaluateGraphQLResolver.js
```diff
@@ -18,17 +20,30 @@ export default function evaluateGraphQLResolver(
...
+    });
+  }
+  return defaultFieldResolver(
+    testValue,
+    argsValue ?? {},
+    contextValue,
+    ({
+      fieldName: field.name,
+      path: { key: 'mocked' },
+    }: any),
```
mrtnzlml, line 47: Would it be possible to sync these params with the direct `field.resolve` on line 36?

src/test-utils/package.json
michalsanger: fixed

src/test-utils/src/evaluateGraphQLResolver.js
michalsanger: I tried

PR #620, opened 2020-04-29
Upgrade relay to 9.1.0

tbergq:
Relay now ships with flow-types. I suppressed the errors with the flow-tool. We can start fixing flow errors after this is merged.

### DIFF
--- a/src/example-relay/src/Homepage/locations/__tests__/LocationsPaginatedRefetch.test.js
+++ b/src/example-relay/src/Homepage/locations/__tests__/LocationsPaginatedRefetch.test.js
@@ -15,6 +15,7 @@ beforeEach(() => {
 
 const TestRenderer = () => (
   <QueryRenderer
+    // $FlowFixMe errors after upgrading to relay 9.1.0
     environment={environment}
     query={graphql`
       query LocationsPaginatedRefetchTestQuery @relay_test_operation {
--- a/src/example-relay/src/LocalForm/index.js
+++ b/src/example-relay/src/LocalForm/index.js
@@ -94,7 +94,11 @@ export default function LocalForm() {
         environment={environment}
         query={graphql`
           query LocalFormQuery {
-            __typename # https://github.com/facebook/relay/issues/2471
+            # https://github.com/facebook/relay/issues/2471
+            # Should be ok, we are not calling the BE anyway
+            locations(first: 0) {
+              __typename
+            }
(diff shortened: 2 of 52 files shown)


src/example-relay/src/LocalForm/index.js
```diff
@@ -94,7 +94,10 @@ export default function LocalForm() {
         environment={environment}
         query={graphql`
           query LocalFormQuery {
-            __typename # https://github.com/facebook/relay/issues/2471
+            # Should be ok, we are not calling the BE anyway
```
mrtnzlml, line 97: The comment `https://github.com/facebook/relay/issues/2471` should stay there - it's not related to this change. Also see: https://github.com/facebook/relay/commit/793729e7af9c7ee0de971e3d2ed26e5896774640#commitcomment-37652508

src/example-relay/src/Homepage/locations/__tests__/LocationsPaginatedRefetch.test.js
```diff
@@ -15,6 +15,7 @@ beforeEach(() => {
 
 const TestRenderer = () => (
   <QueryRenderer
+    // $FlowFixMe upgrading relay
```
mrtnzlml, line 18: Would it be possible to use some more descriptive comment? It would be nice to at least know it's happening from 9.1.

src/example-relay/src/LocalForm/index.js
tbergq: Comment is back in 🙂

src/example-relay/src/Homepage/locations/__tests__/LocationsPaginatedRefetch.test.js
tbergq: Changed to `// $FlowFixMe errors after upgrading to relay 9.1.0`

tbergq:
> Yup, makes sense. Thanks! 👍
> 
> Could you please create an issue to clean up the codebase?

#627

src/example-relay/src/Homepage/locations/__tests__/LocationsPaginatedRefetch.test.js
mrtnzlml: Yes, much better. Now, we can find the exact suppression comment related to this version change easily (and fix it).

PR #628, opened 2020-04-30
relay: Use Environment type from react-relay

tbergq:


src/example-relay/src/SSRQueryRenderer.js
```diff
@@ -31,6 +31,7 @@ export default function SSRQueryRenderer(props: Props) {
 
     return (
       <RelayEnvironmentProvider environment={environment}>
+        {/* // $FlowFixMe errors after upgrading to relay 9.1.0 */}
```
mrtnzlml, line 34: What is the new issue here? 🤔

tbergq: `Cannot call props.onResponse with data bound to the first parameter because inexact SelectorData [1] is incompatible
with exact RendererProps [2]`

I tried fixing it, but it just uncovered other errors, so I thought, better keep it small and take it step by step.

PR #646, opened 2020-05-06
relay-example: Query typename on rootquery in LocalForm

tbergq:


tbergq:
https://github.com/facebook/relay/issues/2471#issuecomment-624238501

tbergq:
> It's a nice workaround, much better. :)

Agreed 🙂

PR #647, opened 2020-05-06
relay: Fix query renderer flow types

tbergq:


src/relay/src/QueryRenderer.js
```diff
@@ -46,6 +43,18 @@ type Props =
       +render: ReadyState => React.Node,
     |};
 
+const getPassOnProps = (props: Props) => {
```
mrtnzlml, line 46: Can you please elaborate on what was wrong here? I am not sure I fully understand this change.

tbergq: Yes, flow gives error since our Props includes these props that is not recognized by relay

```
{|
      ...CommonProps,
      +onSystemError?: ({
        error: Error,
        retry: ?() => void,
        ...
      }) => React.Node,
      +onLoading?: () => React.Node,
      +onResponse: RendererProps => React.Node,
    |}
```

So this is just extracting the prop types that relay expects. I can add a comment in the code

mrtnzlml: I see, yeah, it makes sense. Personally, I'd be more explicit here and removed the spread completely (manually enumerate only the props we are actually using).

tbergq: Ok, sure, that makes sense, I will update with this before merging

tbergq: 🚀

PR #655, opened 2020-05-11
relay: Fix createRelayStore flow types

tbergq:


src/relay/src/createRelayStore.js
```diff
@@ -2,15 +2,17 @@
...
 import { RecordSource, Store } from 'relay-runtime';
 
-import type { RecordMap } from './runtimeTypes.flow';
+import type { RecordMap, Scheduler, OperationLoader, GetDataID } from './runtimeTypes.flow';
 
 type Options = {|
-  +gcReleaseBufferSize?: ?number,
+  gcScheduler?: ?Scheduler,
+  operationLoader?: ?OperationLoader,
+  UNSTABLE_DO_NOT_USE_getDataID?: ?GetDataID,
```
mrtnzlml, line 10: Can we hide this property or it would cause some Flow errors somewhere else?

tbergq: It would cause flow-error yes.

PR #677, opened 2020-05-18
relay: Fix createRequestHandler flow types

tbergq:
- Fix createEnvironment flow types

### DIFF
--- a/src/relay/src/createRequestHandler.js
+++ b/src/relay/src/createRequestHandler.js
@@ -1,18 +1,9 @@
 // @flow
 
-import { Observable } from 'relay-runtime';
-import type { Variables } from '@adeira/relay-runtime';
+import { Observable, type CacheConfig } from 'relay-runtime';
+import type { Variables, GraphQLResponse } from '@adeira/relay-runtime';
 
 import type { RequestNode, Uploadables } from './types.flow';
-
-export type CacheConfig = {|
-  +force?: ?boolean,
-|};
-
-type GraphQLResponse = {|
-  +data?: null | { [key: string]: any, ... },
-  +errors?: $ReadOnlyArray<{ [key: string]: any, ... }>,
-|};
 
 /**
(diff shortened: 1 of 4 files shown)


src/relay/src/createRequestHandler.js
```diff
@@ -1,17 +1,13 @@
...
 import type { RequestNode, Uploadables } from './types.flow';
 
-export type CacheConfig = {|
-  +force?: ?boolean,
-|};
-
 type GraphQLResponse = {|
-  +data?: null | { [key: string]: any, ... },
-  +errors?: $ReadOnlyArray<{ [key: string]: any, ... }>,
+  +data: PayloadData,
```
mrtnzlml, line 9: Property `data` doesn't have to be in response. Is this type correct? 🤔

tbergq: This is the type from `node_modules/relay-runtime/network/RelayNetworkTypes.js.flow`
```
export type GraphQLResponseWithData = {|
  +data: PayloadData,
  +errors?: Array<PayloadError>,
  +extensions?: PayloadExtensions,
  +label?: string,
  +path?: Array<string | number>,
|};
```

mrtnzlml: I can see it's a bit more complicated:

```js
export type GraphQLResponse =
  | GraphQLSingularResponse
  | $ReadOnlyArray<GraphQLSingularResponse>;
```

https://github.com/facebook/relay/blob/26a4fc646cf9a2614ae95468efc90a619bb6c57b/packages/relay-runtime/network/RelayNetworkTypes.js#L44-L84

We already have this type somewhere actually. Maybe we could reuse it?

tbergq: Yeah, that seems to work 👍 I will use it

tbergq: 🚀

PR #681, opened 2020-05-20
Replace read inline

tbergq:


src/relay/src/__flowtests__/readInlineData.js
```diff
@@ -1,94 +0,0 @@
-// @flow
```
mrtnzlml, line 1: Why is this test not valid anymore? It should still work, no?

tbergq: Do we really need it though? We are just exporting the type from `react-relay`

mrtnzlml: It's there to be sure it still works as expected. How do we know the types still work now?

tbergq: If they don't work, how will we fix them?

mrtnzlml: We add the `readInlineData` wrapper back just like we did for years. 🙂

PR #714, opened 2020-05-28
Release @adeira/babel-preset-adeira 0.4.0

tbergq:


tbergq:
I noticed #711 failed, so I want to merge that one first

PR #716, opened 2020-05-29
eslint-runner: Improve lint all logic

tbergq:


mrtnzlml:
@tbergq Can you please provide a commit description with some explanation? Specifically, I am interested in what's the motivation and how it affects `eslintConfig` field in `package.json` since that was the main motivation as far as I know (to catch changes there and lint all).

tbergq:
I didn't think about the `eslintConfig` field in package.json, thanks for pointing it out. 

I thought the motivation for checking package.json was because if you remove a dependency, you need to lint not only touched files to verify the dependency is no longer used in your code. It feels unnecessary to lint all files when upgrading a dependency. 

I will see if I can expand it to check for changes in the `eslintConfig` field as well.

tbergq:
Added check to see if `eslintConfig` changed and added description

mrtnzlml:
> It feels unnecessary to lint all files when upgrading a dependency.

What if you upgrade `eslint-plugin-*` dependency which adds new rules?

I think we should change the behavior of "lint all" a little bit and be more explicit about the intentions. Specifically, we should take into account the `eslintConfig` and _only_ `eslint-plugin-*` dependencies. What do you think? Does it make sense?

mrtnzlml:
> because if you remove a dependency, you need to lint not only touched files to verify the dependency is no longer used in your code

I never thought about this case, to be honest. Can you please elaborate? Wouldn't Flow choke on such code? 🤔

tbergq:
> What if you upgrade eslint-plugin-* dependency which adds new rules?
Good point!

> I never thought about this case, to be honest. Can you please elaborate? Wouldn't Flow choke on such code? 🤔

I saw this happen in the past, before we did lint all on package.json change, and that is why I added package.json here. I guess flow will also report error, if you have flow-types yes. Then, if you remove a package that does not have flow-types, then flow will not catch it. But the `import/no-extraneous-dependencies` will catch it.

src/eslint-runner/src/__tests__/shouldLintAll.test.js
```diff
@@ -10,14 +21,99 @@ test.each([
...
+    }),
+  }));
+  jest.doMock('../../../../package.json', () => ({
+    dependencies: { a: 'lol', b: 'lol2', c: 'lol3' },
+  }));
+
+  expect(shouldLintAll('package.json')).toBe(false);
+});
+
+it('should not lint all for package.json if dependency is upgraded', () => {
```
jaroslav-kubicek, line 54: is this correct? what if I upgrade eslint? 🤔

tbergq: You can do that without even changing package.json

jaroslav-kubicek: that's true (e.g. only yarn.lock can be upgraded and some people might not use it), but still... upgrade in one eslint plugin can produce tons of issues which are discovered much later

tbergq: Yes, one way of diffing, is as @mrtnzlml also mentioned, check if `eslint-plugin-*` is changed. 

I don't think there is a 100% fool-proof way. Even today, you can upgrade eslint and not have the lint runner run lint any files so 🤷‍♂️

jaroslav-kubicek: if it's not an extra complication, I would add this behind some flag or a possibility to opt-out

tbergq: Yeah, I think I will just close this one. I think I can live with how it works today as well.

PR #721, opened 2020-06-02
Upgrade flow to 0.126.1

tbergq:


src/example-relay/src/SSRQueryRenderer.js
```diff
@@ -7,16 +7,14 @@ import { getDataFromRequest } from '@adeira/relay-utils';
 
 import createRelayEnvironment from './createRelayEnvironment';
 
-type RendererProps = {| +[key: string]: any |};
-
-type Props = {|
+type Props<T> = {|
   +query: GraphQLTaggedNode,
   +variables: { [name: string]: mixed, ... },
```
mrtnzlml, line 12: Should we annotate the rest with this type as well? For example, we have something like this:

```ts
readonly render: (relayProps: TOperation['response']) => React.ReactNode;
readonly variables?: TOperation['variables'];
```

So for example, that correct variables are being taken into account.

tbergq: Yeah, I can have a look at it and see 👍

tbergq: I just imported type `Variables` type fro `@adeira/relay-runtime` here.

mrtnzlml: Mm, that's not what I meant because it won't actually check the variables. You should take the actual `<T>` into account in my opinion. But OK - can be improved later if someone is interested in better types.

PR #739, opened 2020-06-08
Hard-code all values in prettier & Release @adeira/eslint-config 2.2.1 

sampi:
- Hard-code all values in prettier
  - to make it clear, these are the DEFAULT values that `@adeira/eslint-config` was using in `v2.1.0`
  - `arrowParens` changed its default value in `prettier@2`, so we should hard-code all values to make sure there are no conflicts going forward.
- Release @adeira/eslint-config 2.2.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "dependencies": {},
   "devDependencies": {
     "@adeira/babel-preset-adeira": "^0.4.0",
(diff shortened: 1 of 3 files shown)


tbergq:
I am closing this, since we released a 2.2.1 where prettier was reverted to 1.19.1, and we released v 3.0.0 with prettier 2.0.5.

PR #742, opened 2020-06-09
Release @adeira/eslint-config 3.0.0

tbergq:


mrtnzlml:
@tbergq Have you considered keeping some settings from v1 like this one for example? https://github.com/adeira/universe/pull/739/files#diff-208aaabc755418e0262022fdc5b46982R51

tbergq:
> @tbergq Have you considered keeping some settings from v1 like this one for example? https://github.com/adeira/universe/pull/739/files#diff-208aaabc755418e0262022fdc5b46982R51

Personally, I like having this parenthesis in my arrow functions. Though, if you feel it is better without it, we can add it

sampi:
@tbergq @mrtnzlml so instead of going with the current `v1` setting you decided to  go with the `v2` default?

I mean, on one hand that's fine because it's the `prettier` default, but on the other hand that's going to cause quite a lot of changed lines and annoyances merging existing/older MRs.

I would recommend going with the `v1` version personally, it also keeps the code looking leaner and easier to read (we're getting into personal opinion territory).

tbergq:
> we're getting into personal opinion territory

Yes, this is exactly it. We could always discuss this more internally, but the biggest changes seems to be: 

- space after anonymous functions:
-- personally I don't like this one, but there is no setting for it (I think)
- parenthesis to arrow function:
-- I like this one, and I think it makes the core more readable IMHO
- Space after inexact object `...` has been removed:
-- I like this one as well. The reason for trailing commas was to avoid the line to appear as changed when doing code review when adding some value to an object. For flow types, the triple dots indicating inexact object have to be the last one, so the trailing comma serves no purpose here. 

These are my personal opinions 😊

PR #743, opened 2020-06-09
Use adeira graphql relay in example-relay

tbergq:
- Release `@adeira/graphql-relay@0.1.0`

### DIFF
--- a/src/example-relay/src/graphql/services/connectionFromArray.js
+++ b/src/example-relay/src/graphql/services/connectionFromArray.js
@@ -6,18 +6,17 @@ import {
   offsetToCursor,
   type Connection,
   type ConnectionArguments,
-} from 'graphql-relay';
+} from '@adeira/graphql-relay';
 
-type Args = $ReadOnly<{
+type Args = {
   ...ConnectionArguments,
   ...
-}>;
+};
 // There is open discusion how to handle bidirectional pagination:
 // https://github.com/graphql/graphql-relay-js/issues/58
 //
 // Until Relay devs decide, let's try our implementation
 export default function connectionFromArray<T>(data: $ReadOnlyArray<T>, args: Args): Connection<T> {
-  // $FlowExpectedError: graphql-relay is not using explicit inexact object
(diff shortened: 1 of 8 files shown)


src/example-relay/src/graphql/services/connectionFromArray.js
```diff
@@ -6,18 +6,17 @@ import {
...
+type Args = {
   ...ConnectionArguments,
   ...
-}>;
+};
 // There is open discusion how to handle bidirectional pagination:
 // https://github.com/graphql/graphql-relay-js/issues/58
 //
 // Until Relay devs decide, let's try our implementation
 export default function connectionFromArray<T>(data: $ReadOnlyArray<T>, args: Args): Connection<T> {
```
mrtnzlml, line 19: I believe we should completely get rid of this function now, right? Can you please create an issue?

yarn.lock
```diff
@@ -6618,13 +6618,6 @@ graphql-extensions@^0.12.2:
     apollo-server-env "^2.4.4"
     apollo-server-types "^0.5.0"
 
-graphql-relay@^0.6.0:
-  version "0.6.0"
-  resolved "https://registry.yarnpkg.com/graphql-relay/-/graphql-relay-0.6.0.tgz#18ec36b772cfcb3dbb9bd369c3f8004cf42c7b93"
-  integrity sha512-OVDi6C9/qOT542Q3KxZdXja3NrDvqzbihn1B44PH8P/c5s0Q90RyQwT6guhGqXqbYEH6zbeLJWjQqiYvcg2vVw==
-  dependencies:
-    prettier "^1.16.0"
```
mrtnzlml, line 6626: Oh, it was another dep which was blocking Prettier. :)

tbergq: Yes, it is still `flow-typed`, we haven't removed it

src/example-relay/src/graphql/services/connectionFromArray.js
tbergq: Yes, that is the plan, I can create the task 👍

PR #768, opened 2020-06-19
Global ID: improve README.md

mrtnzlml:


src/graphql-global-id/README.md
```diff
@@ -50,9 +50,13 @@ const fields = {
 };
 ```
 
-# `toGlobalId`
+# Creating opaque ID imperatively (`toGlobalId`)
 
-TKTK
+You can you `toGlobalId` to create the unique and opaque identifier manually. However, try to always use the `GlobalID` type constructor since it handles the type automatically. It's usually a bad practice to use `toGlobalId` in your types unles you really need it. Usage:
```
tbergq, line 55: `You can you `toGlobalId`` -> `You can use`?

tbergq: `unles` -> `unless`

mrtnzlml: thanks

src/graphql-global-id/README.md
```diff
@@ -86,3 +90,5 @@ It is possible to fetch both ID versions at the same time in GraphQL:
 ```
 
 This is handy especially when you are migrating old code to this new type. Just change something like deprecated `databaseID` to `databaseID: id(opaque: false)` and that's it.
+
+Generally speaking, it's a bad practice to fetch `id(opaque: false)` without aliasing it! It's because it could lead to unexpected side effects in consistency of your UI. Always use the opaque ID for cache key (never the non-opaque version). Remember, the non-opaque ID is not unique.
```
michalsanger, line 94: The last sentence maybe:
Remember, the non-opaque ID is not guaranteed to be globally unique.

PR #769, opened 2020-06-20
Improve glob options flow type

tbergq:
Release @adeira/monorepo-utils 0.8.0


tbergq:
Copy pasted and adapted the typescript types. So it should be ok 😊

PR #816, opened 2020-07-01
Release @adeira/js version 1.2.3

mrtnzlml:


tbergq:
What is new, should we update the change log?

mrtnzlml:
@tbergq Not necessary in my opinion. There are some insignificant maintenance changes and `@babel/runtime` added into dependencies (that's what I want to release). Dunno why it's not released already (hence the patch). 🤔

PR #834, opened 2020-07-14
Upgrade relay to 10.0.0

tbergq:


src/relay-runtime/src/RelayLogger.js
```diff
@@ -82,6 +82,13 @@ export default function RelayLogger(logEvent: LogEvent) {
     case 'execute.unsubscribe':
       logGroup(logEvent);
       break;
+    // New events, maybe add log function later
+    case 'store.notify.complete':
+    case 'store.notify.start':
+    case 'store.gc':
+    case 'store.restore':
+    case 'store.snapshot':
+    case 'store.publish': // End new events
```
mrtnzlml, line 91: Feel free to remove the comments - I don't think they add much value.

tbergq:
> Nice, thanks! 👍
> 
> Have you tried to run it? Is it without any issues?

Yeah, I ran the example relay locally and it worked fine

src/relay-runtime/src/RelayLogger.js
tbergq: Ok, will do, I just wanted to point out that we might want to add some relay logger there, but I agree, I can remove the comment 👍

PR #886, opened 2020-07-31
js: Check getOwnPropertyNames also to determine if object is empty

tbergq:


scripts/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.0.0",
   "license": "UNLICENSED",
   "dependencies": {
-    "@adeira/js": "^1.2.0",
+    "@adeira/js": "^1.2.4",
```
jaroslav-kubicek, line 7: might be worth adding some tests for this -> `yarn workspaces info` => dependency is added to `mismatchedWorkspaceDependencies` in such cases

tbergq: Feel free to create a PR for it 🙂

tbergq:
> Looks fine, do you have some use case? 🤔

The use case would be every time you use the function. If we really want to check if the object is empty(has no properties and extends Object) then we should really check that it is. It is probably edge case for an object to have only non-enumerable properties, but it is certainly not _empty_.

PR #904, opened 2020-08-07
Dependencies: remove Eslint resolution

mrtnzlml:
I tested it without it and it seems to be working fine. I don't even remember why was it there. Seems like we no longer have this problem (?).


tbergq:
Can you try running `yarn install --force`, I would expect `yarn.lock` to change also 🤔

mrtnzlml:
@tbergq Updated.

tbergq:
Thanks, also, shouldn't pipeline fail in these cases, where yarn.lock is out of date?

mrtnzlml:
Yes, but I am not sure if this is really being out of date since there was no effect on the installed versions. Dunno.

tbergq:
> since there was no effect on the installed versions

Ok, that sounds reasonable 👍

PR #909, opened 2020-08-10
Fetch: implement Abort Controller (cancellable fetch)

mrtnzlml:
Test plan: start Relay Example, slow down your network request and click furiously - the identical network requests should be cancelled (best visible in bi-directional pagination example).

### DIFF
--- a/src/fetch/README.md
+++ b/src/fetch/README.md
@@ -1,7 +1,8 @@
-This package has been extracted from the original [fbjs](https://github.com/facebook/fbjs/blob/master/packages/fbjs/src/fetch/fetchWithRetries.js) library and it exposes single `fetchWithRetries`. This function is only a wrapper for any other well known [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). However, this fetch also solves two common problems:
+This package has been extracted from the original [fbjs](https://github.com/facebook/fbjs/blob/master/packages/fbjs/src/fetch/fetchWithRetries.js) library and it exposes single `fetchWithRetries`. This function is only a wrapper for any other well known [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). However, this fetch also solves these common problems:
 
-- fetch timeouts, and
-- retries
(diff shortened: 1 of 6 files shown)


src/fetch/README.md
```diff
@@ -1,7 +1,8 @@
-This package has been extracted from the original [fbjs](https://github.com/facebook/fbjs/blob/master/packages/fbjs/src/fetch/fetchWithRetries.js) library and it exposes single `fetchWithRetries`. This function is only a wrapper for any other well known [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). However, this fetch also solves two common problems:
+This package has been extracted from the original [fbjs](https://github.com/facebook/fbjs/blob/master/packages/fbjs/src/fetch/fetchWithRetries.js) library and it exposes single `fetchWithRetries`. This function is only a wrapper for any other well known [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). However, this fetch also solves these common problems:
 
-- fetch timeouts, and
-- retries
+- fetch timeouts
+- retries, and
+- request cancellations
```
mrtnzlml, line 5: There is technically no change in the Fetch realm but I thought mentioning would be beneficial.

src/relay/src/createRequestHandler.js
```diff
@@ -18,18 +19,17 @@ type Sink = {|
...
     requestNode: RequestNode,
     variables: Variables,
     cacheConfig: CacheConfig,
     uploadables: ?Uploadables,
   ) {
+    const controller = new AbortController();
+    const signal = controller.signal;
+
     return Observable.create((sink: Sink) => {
-      customFetcher(requestNode, variables, uploadables)
```
tbergq, line 32: Should we add a note in the `relay/readme.md` about how to use the fetch abort function

mrtnzlml: I'd skip it for now. We are explicitly not recommending to write custom fetchers but rather use the one which we provide and that one solves it by design.

PR #910, opened 2020-08-10
Add note about msw to readme

tbergq:
This is an alternative to mocking the fetch function

### DIFF
--- a/src/fetch/README.md
+++ b/src/fetch/README.md
@@ -165,3 +165,13 @@ it('mocks the fetch', async () => {
 ```
 
 Why mocking `global.fetch` doesn't work? It's because this fetch doesn't use or pollute `global` variable: it uses ponyfill instead of polyfill behind the scenes.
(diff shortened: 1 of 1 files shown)


src/fetch/README.md
```diff
@@ -165,3 +165,13 @@ it('mocks the fetch', async () => {
 ```
 
 Why mocking `global.fetch` doesn't work? It's because this fetch doesn't use or pollute `global` variable: it uses ponyfill instead of polyfill behind the scenes.
+
+
+## Msw, an alternative to mock fetch
```
mrtnzlml, line 170: Is this a good place where to put it? Under "Frequently Asked Questions"? 🤔 It's not a question. Maybe `###` as a subsection to the answer above?

tbergq: Yeah, you are right, subsection is better here, it belongs under the mock section.

PR #911, opened 2020-08-10
Apply Prettier to markdown files

mrtnzlml:
Currently, we are checking only JS files via Eslint. This change adds additional check to markdown files as well.

### DIFF
--- a/src/babel-preset-adeira/CHANGELOG.md
+++ b/src/babel-preset-adeira/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 Changelog before our fork:
 
-```text
+<pre><code>
 # Unreleased
 
 - Output in which file you are trying to redeclare `__DEV__`
@@ -114,4 +114,4 @@ module.exports = {
 # 1.1.0
 
 - Added support for `__DEV__` expression
-```
+</code></pre>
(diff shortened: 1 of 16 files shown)


src/babel-preset-adeira/CHANGELOG.md
```diff
@@ -55,12 +55,12 @@ Changelog before our fork:
 - Added support for dynamic `import()` syntax. This fits well with `@adeira/relay` version 1.0+ and allows you to use `@match`/`@module` directives for data-driven code splitting.
 - Breaking: Relay plugin is no longer part of this preset. It was moved to `@adeira/relay`. Please, edit your Babel configuration files (example for Next.js applications):
 
-```js
+\`\`\`js
```
michalsanger, line 58: This does not look OK.

mrtnzlml: How should we do it differently? We are nesting there one code block into another, see: https://github.com/adeira/universe/blob/5d3733e6ea9c04c32e2d434d394b0081bdb0afe3/src/babel-preset-adeira/CHANGELOG.md

michalsanger: Options I see:

1. close the block before the one with js example
2. remove the the triple backtick
3. don't format the whole "Changelog before our fork" as a code block

mrtnzlml: None of the aforementioned options are what I wanted to do. But it's fixed now, please have a look.

Issue #898, opened 2020-08-05
Fetch: add cancelation feature


mrtnzlml:
See: https://developer.mozilla.org/en-US/docs/Web/API/AbortController

Re-revert this PR with native `AbortController` if it exists: https://github.com/adeira/universe/pull/916 (ignore the cancelation otherwise). Please note: we also have to handle `AbortError` in our fetch so fetch do not retry in this case.

- https://developer.mozilla.org/en-US/docs/Web/API/AbortController
- https://developers.google.com/web/updates/2017/09/abortable-fetch

mrtnzlml:
Happy birthday! 🥳 

I would like to close this issue as it is being stale and we were not able to fix it in a reasonable time. It could be due to insufficient resources from our side or we simply do not consider it to be a big deal.

Here are some options what to do now:

1. You can send a pull request if you consider the issue to be important for you and you'd like to contribute back. ✍️ 
1. You can consider sponsoring one of our maintainers to help us deal with the associated maintenance and development costs. It will be greatly appreciated. 👍 
1. You can fork our open-source projects and start developing your own version. Consider contributing back first though.
1. Do nothing.

Thanks for understanding. ❤️

PR #916, opened 2020-08-11
Relay: partially revert abort controller

mrtnzlml:
This partially reverts 5acc6c1d5a5eb0eb8e09c902b5d3ba3b92b28ec7 - turned out it's a bit tricky to implement it properly to support even IE 11. Will try again somewhere down the road.

### DIFF
--- a/src/relay/CHANGELOG.md
+++ b/src/relay/CHANGELOG.md
@@ -1,7 +1,5 @@
 # Unreleased
 
-- Automatically cancel stale requests via fetch `AbortController` (https://developer.mozilla.org/en-US/docs/Web/API/AbortController).
-
 # 2.0.2
 
 - Upgrade relay to 10.0.1
--- a/src/relay/package.json
+++ b/src/relay/package.json
@@ -25,7 +25,6 @@
     "@adeira/signed-source": "^1.0.0",
     "@babel/register": "^7.10.5",
     "@babel/runtime": "^7.11.2",
-    "abort-controller": "^3.0.0",
     "babel-plugin-relay": "^10.0.1",
     "commander": "^6.0.0",
     "is-ci": "^2.0.0",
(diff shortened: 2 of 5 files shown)


michalsanger:
We do care about IE 11?

tbergq:
> We do care about IE 11?

In kiwi, we still do, and we use this package

mrtnzlml:
@michalsanger Kiwi.com apparently does so we have to back up a little bit. :)

michalsanger:
How about a new major version?

mrtnzlml:
@michalsanger But that would essentially mean that Kiwi.com would have to stop using this package which is a pity and I'd say unwanted.

tbergq:
But it should be usable if you import it like `import AbortController from 'abort-controller/dist/abort-controller.umd'`

mrtnzlml:
@tbergq But then there is the problem with fetch retries because of "AbortError" and I was not able to solve it via the `abort-controller` package. I am thinking that we could introduce a new fetcher `createAbortableNetworkFetcher.js` if anyone really wants it. 🤔

tbergq:
But I told you I was able to run it in IE-11 with the import described above 🤔

mrtnzlml:
@tbergq Yes, me too thanks to your advice. But what about the abort error retries you mentioned?

Screenshot 2020-08-11 at 12 45 57

Anyway, Michal had a good idea now that we could support it only when it's natively available in the browser and ignore it otherwise. I like it. Still, let's revert this PR so we are not blocking future versions and I will create a followup issue to re-revert it with this change. What do you think?

tbergq:
> @tbergq Yes, me too thanks to your advice. But what about the abort error retries you mentioned?

Sorry, I guess I should have been more specific 🤦  

The only thing wrong with that image is that it logs fetchWithRetries: Aborted twice. I thought it was retrying the request, but I think it is not. Since there were no more requests in the network inspector. 

The DOMExcpetion below is the console.log I added in fetch().catch block

mrtnzlml:
> I thought it was retrying the request, but I think it is not. Since there were no more requests in the network inspector.

That sounds even worse because when you look at the `fetchWithRetries` code, it actually tries to retry it (which I guess cannot but it tries).

mrtnzlml:
Reverting for now but I am leaving this issue open to re-implement it again with native `AbortController` if it exists: https://github.com/adeira/universe/issues/898

PR #922, opened 2020-08-14
Global ID: Export Encoder

michalsanger:
I'm gonna need it for `graphql-relay-fauna` package


src/graphql-global-id/src/index.js
```diff
@@ -0,0 +1,8 @@
+// @flow srtict
```
mrtnzlml, line 1:
```suggestion
// @flow strict
```

We should update our Eslint rule to catch invalid values. Could you please create an issue and link this PR?

PR #926, opened 2020-08-15
Eslint-config: Check valid flow modes

### LINKED ISSUE #925: @adeira/eslint-plugin: Add rule to check valid Flow mode values
Currently there is no typo check for the Flow flag. Only these values should be allowed:
- `@flow`
- `@flow strict`
- `@flow strict-local`

It would help in #922

tbergq:
closes #925


mrtnzlml:
@tbergq That was fast! 😊 I added a little improvement which I was working on locally: https://github.com/adeira/universe/pull/928

PR #930, opened 2020-08-16
Flowconfig: test 'well_formed_exports.includes' paths

mrtnzlml:
This change adds a test which verifies that all 'well_formed_exports.includes' paths are valid and enables more projects with well formed exports.


src/monorepo-shipit/config/docs-adeira-dev.js
```diff
@@ -1,7 +1,7 @@
 // @flow strict
 
 module.exports = {
-  getStaticConfig() {
+  getStaticConfig(): { repository: string, ... } {
```
michalsanger, line 4: Why not `{| repository: string |}`?

mrtnzlml: Changed. It's autogenerated code and I didn't touch it unless it's a total disaster. :)

PR #940, opened 2020-08-18
Global ID: fix TS types declarations

mrtnzlml:
The package structure changed in 5789f490f93e650bc9f3701374940960b6f94186 but the TS types were not in sync. This change fixes it and makes it work on TS projects again.


mrtnzlml:
cc @michalsanger

PR #943, opened 2020-08-18
Flow: enable types first architecture

mrtnzlml:
See: https://medium.com/flow-type/types-first-a-scalable-new-architecture-for-flow-3d8c7ba1d4eb

I used `$FlowFixMe` in many cases, however, it's quite often in an export boundary of Flow tests which we don't need to type precisely (they are never going to be imported anywhere).


src/example-relay/src/SSRLocations/LocationsQuery.js
```diff
@@ -8,7 +8,7 @@ import QueryRenderer from '../SSRQueryRenderer';
 import type { LocationsQueryResponse } from './__generated__/LocationsQuery.graphql';
 import LocationsList from './LocationsList';
 
-export const query = graphql`
+export const query: $FlowFixMe = graphql`
```
tbergq, line 11: This is `GraphQLTaggedNode` type exported from `@adeira/relay`

tbergq (approved):

PR #980, opened 2020-08-26
SX: Fix example in Readme

michalsanger:
Flow type from Readme example removed to make it more simple.

Added import from `next/document` to make the example copy&paste ready.

### DIFF
--- a/src/sx/README.md
+++ b/src/sx/README.md
@@ -56,6 +56,7 @@ Finally, render somewhere the styles and HTML. Example for Next.js with [custom 
 
 ```jsx
 import * as sx from '@adeira/sx';
+import Document, { Html, Head, Main, NextScript } from 'next/document';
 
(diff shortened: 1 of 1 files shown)


src/sx/README.md
```diff
@@ -61,17 +61,17 @@ export default class MyDocument extends Document {
     return { ...html, css };
   }
 
-  render(): React.Node {
+  render() {
     return (
-      <Html>
+      <html>
```
tbergq, line 66: I think this should stay as `Html` [next-custom-document](https://nextjs.org/docs/advanced-features/custom-document)

michalsanger: I see, so add import from `next/document` maybe?

mrtnzlml: Could be, yes - it's a Next.js example so it makes sense.

mrtnzlml (approved):

PR #984, opened 2020-08-28
sx: Refactor create function

tbergq:
- Support pseudo classes in media queries

### DIFF
--- a/src/example-relay/package.json
+++ b/src/example-relay/package.json
@@ -5,7 +5,7 @@
   "private": true,
   "dependencies": {
     "@adeira/fetch": "^1.0.3",
-    "@adeira/sx": "^0.5.0",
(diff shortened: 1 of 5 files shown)


mrtnzlml (approved):
Nice, thanks! We are slowly iterating forward and I like it. 😊 

Would it be possible to add some info to our readme? Specifically, we have there this:

> _Note: we support only one level of media queries at this moment, they cannot be nested._

Is it still valid?

tbergq:
> Would it be possible to add some info to our readme? Specifically, we have there this:

> Note: we support only one level of media queries at this moment, they cannot be nested.

> Is it still valid?

Good point, I will update it

mrtnzlml (approved):

PR #986, opened 2020-08-28
SX: Suggest solution in error message

michalsanger:
The error message when an unknown sheet name is used can be too long if there are a lot of sheet definitions.

A better DX could be just to suggest an alternative in the error message instead of a list of all available sheet names.


tbergq (approved):

mrtnzlml (approved):
Thanks, I love these suggestions! 👍

PR #987, opened 2020-08-29
example-relay: Make it responsive

tbergq:


src/sx/src/css-properties/all-properties.en.json
```diff
@@ -4049,6 +4049,12 @@
     "status": "CR",
     "title": "CSS Grid Layout Module Level 2"
   },
+  {
+    "property": "grid-gap",
+    "url": "Did't find it",
+    "status": "?",
+    "title": "?"
+  },
```
mrtnzlml, line 4057: We should not extend this file manually but rather download it. Seems like we are talking about this property? https://developer.mozilla.org/en-US/docs/Web/CSS/gap

Sounds like a job for some prefixer, see:

> CSS Grid Layout initially defined the grid-gap property. This prefixed property is being replaced by gap. However, in order to support browsers that implemented grid-gap and not gap for grid, you will need to use the prefixed property as in the interactive example above.

tbergq: 🚀

mrtnzlml (approved):

PR #990, opened 2020-08-29
SX: add failing test for nested media queries

mrtnzlml:
We are currently generating media queries on a global level but they should stay nested. See: https://www.w3.org/TR/css3-conditional/#processing


tbergq:
Looks like we need a smarter way to pass down the current style buffer 🙈

mrtnzlml:
@tbergq Or perhaps change the whole structure of it and make it more advanced. It's currently very very simple.

tbergq:
Yeah, that is probably possible as well 👍

mrtnzlml:
Support added in SX version 0.10.0 - closing.

PR #994, opened 2020-08-29
Improve Eslint runner output

mrtnzlml:
This change adds a custom reported to improve the output while 1) showing the progress (so we know it's not stuck) and 2) showing all the warnings (todos) and errors at the end. This was the main motivation because with the previous solution we were not showing the warnings at the end of the lint run.

Screenshot 2020-08-29 at 16 22 22


tbergq (approved):

tbergq:
Nice, looking good 👌

PR #1011, opened 2020-08-30
example: Replace grommet with sx styled components

tbergq:
- Currently deactivate client side routing, since it doesn't work with sx
- Next up: remove gromment-icons, so we can remove styled-components

### DIFF
--- a/src/example-relay/src/components/Button.js
+++ b/src/example-relay/src/components/Button.js
@@ -1,0 +1,73 @@
+// @flow
+
+import * as React from 'react';
+import * as sx from '@adeira/sx';
+
+type Props = {|
+  +onClick?: () => void,
+  +children: React.Node,
+  +dataTest?: string,
+  +disabled?: boolean,
+  +iconLeft?: React.Node,
+  +iconRight?: React.Node,
+  +type?: 'submit' | 'button',
+|};
+
+export default function Button({
+  children,
+  dataTest,
+  iconLeft,
+  iconRight,
+  type = 'button',
+  ...rest
+}: Props): React.Node {
+  return (
+    <button
+      // eslint-disable-next-line react/button-has-type
+      type={type}
+      data-testid={dataTest}
+      className={styles('button', 'primary', rest.disabled ? 'disabled' : null)}
+      {...rest}
+    >
+      <span className={styles('spacer')}>
+        {iconLeft}
+        {children}
+        {iconRight}
+      </span>
+    </button>
+  );
+}
+
+const styles = sx.create({
+  button: {
+    cursor: 'pointer',
+    outline: 'none',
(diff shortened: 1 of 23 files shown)


src/example-relay/src/components/Button.js
```diff
@@ -0,0 +1,73 @@
...
+  iconRight,
+  type = 'button',
+  ...rest
+}: Props): React.Node {
+  return (
+    <button
+      // eslint-disable-next-line react/button-has-type
+      type={type}
+      data-testid={dataTest}
+      className={styles('button', 'primary', rest.disabled ? 'disabled' : 'button')} // TODO: Allow conditional style in sx
```
mrtnzlml, line 29: This should be already possible in the latest SX version. Could you please update this code? Or is there still some problem we need to fix?

tbergq: No, I forgot to fix it after rebasing, thanks for pointing it out 😊

src/example-relay/src/pages/_app.js
```diff
@@ -102,9 +71,21 @@ const styles = sx.create({
...
     maxWidth: '100%',
-    padding: 12,
+    padding: 'var(--space-small)',
   },
   boxWarning: {
     backgroundColor: '#FFCA28',
   },
+  padMedium: {
+    padding: 'var(--space-small)',
+    [tablet]: {
```
mrtnzlml, line 81: This is an interesting technique. I kinda like it. :)

src/example-relay/src/components/Button.js
tbergq: 🚀

src/example-relay/src/pages/_app.js
tbergq: Yeah, it seems like the best way to keep your breakpoints in sync throughout the app 🙂

src/example-relay/src/components/TextInput.js
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <label>
+      <Text size="small">{label}</Text>
+      <input {...rest} className={styles('input')} onChange={handleChange} />
+    </label>
+  );
+}
+
+const styles = sx.create({
+  label: {
```
mrtnzlml, line 29: Is this style unused? 🤔 

We need to figure out some simple lint.

tbergq: Yes, it is, some linting could definitely be added at some point yes 👍

tbergq: 🚀

mrtnzlml (approved):
It's an interesting overview of SX usage. It will definitely help us to develop it further. Good job!

PR #1026, opened 2020-09-01
SX: Some more tests added

michalsanger:
Tests for Flex, Grid and z-index


mrtnzlml:
@michalsanger Could you please share a failing test (of Flow test)? I have this feeling that the number is correctly defined as unitless but the type is not precise. Could that be the case?

michalsanger:
Tests added. I included changes also in `z-index`

src/sx/src/css-properties/isUnitlessNumber.js
```diff
@@ -41,7 +31,6 @@ export default {
   orphans: true,
   tabSize: true,
   widows: true,
-  zIndex: true,
```
mrtnzlml, line 44: Can you please revert all the changes in this file? It should not be changed - all the values there are correct (check the description).

The proper way how to fix this should be to remove this logic here: https://github.com/adeira/universe/blob/a88d1102c5c0c26b97df984ffd5e29534ca922d5/src/sx/src/css-properties/generate.js#L27 (keep there only `string | number` even though it's less precise). It's incorrect to confuse unitless numbers with `number` type. My mistake in the generator. 🙈

src/sx/src/transformValue.js
```diff
@@ -12,6 +12,11 @@ export default function transformValue(
     // https://engineering.fb.com/web/facebook-com-accessibility/
     const defaultFontSize = 16;
     return `${Number(styleValue) / defaultFontSize}rem`;
+  } else if (styleName === 'zIndex') {
+    if (styleValue === 'auto') {
+      return `${styleValue}`;
+    }
+    return `${Number(styleValue)}`;
```
mrtnzlml, line 19: This should also be removed. It's going to work just fine once we fix the generator (unless I am missing something). Of course, the generated Flow types will be a bit more imprecise but that's correct. Luckily, we have a way how to make them more precise again - you can overwrite them here: https://github.com/adeira/universe/blob/2f2fa898f442feda1ec0be225ca2e962c230263f/src/sx/src/css-properties/CustomPropertyTypes.js

mrtnzlml:
@michalsanger I think I am not explaining the underlying problem well so here is the generator fix: https://github.com/adeira/universe/pull/1037 This means that you can revert the changes in `CustomPropertyTypes.js` since that's the default.

mrtnzlml:
And here, I am introducing pretty dope change which will allow us to generate even more precise types: https://github.com/adeira/universe/pull/1041 🥳 

So we could reduce this PR to just tests and flowtests and merge it. :)

mrtnzlml (approved):

PR #1035, opened 2020-09-03
Relay runtime: add "lazy" and "debug" logger

mrtnzlml:
I've seen something similar in one FB talk which make me realize that we are logging unnecessary garbage. So I renamed the current logger to "eager" logger and created a new "lazy" one. The difference is, that this logger waits for the response and prints all the information at once (instead of streaming them into console as they arrive). This is a first iteration but eventually I'd like to use the lazy logger as a default one inside our Relay package.


src/relay-runtime/src/__tests__/RelayLazyLogger.test.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * @flow
+ * @jest-environment jsdom
```
tbergq, line 3: Isn't this already what we use in this workspace?

src/relay-runtime/README.md
```diff
@@ -1,21 +1,29 @@
...
+An opinionated Relay Runtime wrapper with nice query logging.
 
 ## Install
 
 `yarn add @adeira/relay-runtime`
 
 ## Usage
 
-### Logger
+TKTK
```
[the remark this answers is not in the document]
mrtnzlml: yes, would be nice to write some docs but I don't want to do it now 😬

tbergq (approved):

src/relay-runtime/src/__tests__/RelayLazyLogger.test.js
mrtnzlml: It is, however, we should also have a Node.js test which is currently missing. You are the author actually in the eager logger and I like to keep it there explicitly. 😊

tbergq: Sure, makes sense 👌

PR #1041, opened 2020-09-04
SX: improve generated Flow types using 'mdn-data'

mrtnzlml:
This makes some types less accurate because I removed `CustomPropertyTypes` but at the same time it makes many properties more accurate because we actually have a source for these props. We should keep iterating on the generator and effectively make the types more accurate (instead of writing the types manually).


src/sx/src/css-properties/generateFlowTypes.js
```diff
@@ -0,0 +1,136 @@
...
+    return null;
+  }
+}
+
+let flowPrint = '';
+for (const property of allProperties) {
+  const propertyData = mdnData.css.properties[property];
+
+  // default Flow type
+  let flowTypes = new Set([FLOW_TYPE_NUMBER, FLOW_TYPE_STRING]);
```
michalsanger, line 80: I'd move the `new Set` out of this loop, no need to create it over and over again.

src/sx/src/css-properties/generateFlowTypes.js
```diff
@@ -0,0 +1,136 @@
...
+      flowTypes.add('unset');
+    }
+  }
+
+  const propertyName = changeCase.camelCase(property);
+  if (propertyName !== '') {
+    if (flowTypes.has(FLOW_TYPE_STRING)) {
+      // If the type collection yields `string` type, then there is no point in including string
+      // enums in the input (because `string` type would swallow it anyway).
+      flowTypes = new Set(Array.from(flowTypes).filter((type) => typeof type !== 'string'));
```
michalsanger, line 100: Maybe better to read:
```suggestion
      flowTypes = flowTypes.forEach(type => typeof type === 'string' && flowTypes.delete(type));
```

src/sx/src/css-properties/generateFlowTypes.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    const flowType = Array.from(flowTypes)
+      .map((type) => {
+        return typeof type === 'symbol' ? Symbol.keyFor(type) : `"${type}"`;
+      })
+      .join('|');
+
+    const mdnUrlComment =
+      propertyData && propertyData.mdn_url != null ? `// ${propertyData.mdn_url}` : '';
```
michalsanger, line 110:
```suggestion
      propertyData?.mdn_url != null ? `// ${propertyData.mdn_url}` : '';
```

michalsanger (approved):
Shouldn’t the package version be bumped?

src/sx/src/css-properties/generateFlowTypes.js
mrtnzlml: It has to stay there to reset each property type back to `number | string`. Otherwise, I'd have to reset it somewhere at the back of the loop.

PR #1049, opened 2020-09-11
SXTailwind: SX styles generated from Tailwind CSS

michalsanger:


src/tailwind-sx/src/generate.js
```diff
@@ -0,0 +1,36 @@
...
+ * Tailwind CSS file converted into SX stylesheets
+ *
+ * @see https://tailwindcss.com/
+ * @see https://github.com/adeira/sx
+ */
+
+import * as sx from '@adeira/sx';
+
+// eslint-disable-next-line flowtype/no-weak-types
+const tailwindSX: Object = sx.create(${JSON.stringify(styles)});
```
mrtnzlml, line 28: Would it be possible to avoid weak types like `Object` please? Some alternatives are suggested here: https://github.com/adeira/universe/blob/67e496b09730e6ed01a3f727179ff7f6b2d90cef/src/eslint-config-adeira/ourRules.js#L285-L286

src/tailwind-sx/src/sxGenerator.js
```diff
@@ -0,0 +1,98 @@
...
+
+  csstree.walk(ast, {
+    enter(node) {
+      if (node.type === 'ClassSelector') {
+        if (this.atrule != null) {
+          return;
+        }
+
+        const pseudoClassSelector = getPseudoSelectorName(this.rule);
+        const className = node.name.replace('\\', '');
```
mrtnzlml, line 27: This GitHub code scanning warning is worth checking (not sure if it's valid or if you saw it) - see below this comment.

yarn.lock
```diff
@@ -11663,6 +11833,11 @@ prettier@^2.0.5:
   resolved "https://registry.yarnpkg.com/prettier/-/prettier-2.0.5.tgz#d6d56282455243f2f92cc1716692c08aa31522d4"
   integrity sha512-7PtVymN48hGcO4fGjybyBSIWDsLU4H4XlvOHfq91pz9kkGlonzwTfYkaIEwiRg/dAJF9YlbsduBAgtYLi+8cFg==
 
+prettier@^2.1.1:
+  version "2.1.1"
```
mrtnzlml, line 11837: Could you please install the same version of prettier we already use? There are some troubles with this version which prevents us from upgrading and I don't think it's a good idea to have 2 versions.

src/tailwind-sx/src/sxGenerator.js
```diff
@@ -0,0 +1,98 @@
...
+const NOT_SUPPORTED_PROPS = new Set([
+  'grid-gap',
+  'grid-row-gap',
+  'grid-column-gap',
+  'pointer-events',
+]);
+const UNITLESS_PROPS = new Set(['line-height', 'font-size']);
+
+// eslint-disable-next-line flowtype/no-weak-types
+export default function generateSX(css: string | Buffer): Object {
```
mrtnzlml, line 15: Would it be possible to avoid weak types in this file as well, please? We spend great effort migrating the codebase from this legacy type and I don't want to go back (we still allow `any` though). 😬 You can also replace it with `$FlowFixMe` type suppress type like so if you don't want to deal with it now:

```suggestion
export default function generateSX(css: string | Buffer): $FlowFixMe {
```

mrtnzlml (commented):
Looks good, added some minor comments. The `flowtype/no-weak-types` rules would be nice to fix.

mrtnzlml (approved):

PR #1064, opened 2020-09-18
Dependencies: use latest 'flow-bin' version 0.134.0

mrtnzlml:
See: https://github.com/facebook/flow/releases/tag/v0.134.0


tbergq:
types_first should be default now, we can remove it from .flowconfig

mrtnzlml:
@tbergq Updated. Hopefully, it still works 🤞

tbergq (approved):

Issue #1078, opened 2020-09-25
CI: add cache for Flow saved state


mrtnzlml:

mrtnzlml:
Mm, seems like this is currently not possible with Github CI, see: https://github.com/actions/cache/issues/106

PR #1079, opened 2020-09-25
CI: add cache for Flow saved state

mrtnzlml:
Closes: https://github.com/adeira/universe/issues/1078

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -19,10 +19,20 @@ jobs:
 
     steps:
       - uses: actions/checkout@v1
+
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
+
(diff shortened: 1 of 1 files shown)


tbergq (approved):

tbergq:
Actually, I don't think it works to use `hashFiles` on a gitignored file

tbergq:
> Actually, I don't think it works to use `hashFiles` on a gitignored file

Since it says `Cache restored from key: Linux-flow-`, there is no hash there

mrtnzlml:
@tbergq So how should it be? Just without the `hashFiles` and without the `restore-key`?

tbergq:
I really don't know. If the cache key doesn't change, but the file itself changes, the changed file is not uploaded to the remote cache which is not what we want. Maybe artifacts are better here?

mrtnzlml:
Mm, seems like we cannot cache it now on Github CI (https://github.com/actions/cache/issues/106). Closing for now.

PR #1085, opened 2020-09-26
Make flow bin work in repos that are not using yarn workspaces

tbergq:
- Fix crash in `@adeira/monorepo-utils` when there is not git remote
- Release a new version of monorepo utils and flow-bin.


src/flow-bin/bin/index.js
```diff
@@ -27,7 +27,15 @@ for (const command of argv) {
   }
 }
 
-const monorepoRoot = findMonorepoRoot();
+const monorepoRoot = (() => {
+  try {
+    return findMonorepoRoot();
+  } catch {
+    // In case it is not used in a monorepo, fallback to process.cwd
+    return process.cwd();
```
tbergq, line 35: alternatively, we could accept root as a parameter and use it instead of findMonorepoRoot

src/monorepo-utils/src/Git.js
```diff
@@ -52,8 +52,13 @@ const Git = {
   },
 
   getChangedFiles() /*: $ReadOnlyArray<string> */ {
-    const rawChanges = git('diff', '--name-only', 'origin/master...HEAD');
-    return __parseRows(rawChanges);
+    try {
+      const rawChanges = git('diff', '--name-only', 'origin/master...HEAD');
+      return __parseRows(rawChanges);
+    } catch {
+      // This command fails if you have no remote, empty string should be a reasonable fallback
```
mrtnzlml, line 59: Would it make more sense to have `origin/master` configurable (so you can set just `master`) instead of silently ignoring the missing ref?

src/flow-bin/bin/index.js
mrtnzlml: I think this is fine. But would it be possible to DRY it with `src/flow-bin/src/index.js`? Since it's the same package and the logic should be always the same.

tbergq: 🚀

src/monorepo-utils/src/Git.js
tbergq: 🚀

src/monorepo-utils/src/Git.js
```diff
@@ -52,8 +52,14 @@ const Git = {
...
 
   getChangedFiles() /*: $ReadOnlyArray<string> */ {
-    const rawChanges = git('diff', '--name-only', 'origin/master...HEAD');
-    return __parseRows(rawChanges);
+    try {
+      const rawChanges = git('diff', '--name-only', 'origin/master...HEAD');
+      return __parseRows(rawChanges);
+    } catch {
+      // This command above fails if you have no remote, this should work until you add a remote
+      const rawChanges = git('diff', '--name-only', 'master');
```
mrtnzlml, line 60: Shouldn't it be like this?

```suggestion
      const rawChanges = git('diff', '--name-only', 'master...HEAD');
```

tbergq:
@mrtnzlml please check again

mrtnzlml (approved):

PR #1087, opened 2020-09-27
SX: Render media queries as last on the page

michalsanger:
In CSS the rule that is set last, is the one that will be executed. That means rules in media queries would never be used if they are not at the end.

Here the red always wins:
```
@media (max-width: 640px) {
  color: blue;
}
color: red;
```

Here the color is blue on a screen smaller than 640px:
```
color: red;
@media (max-width: 640px) {
  color: blue;
}
```

### DIFF
--- a/src/sx/src/StyleCollector.js
+++ b/src/sx/src/StyleCollector.js
@@ -26,7 +26,7 @@ export class StyleNode {
   }
 }
 
-class StyleCollector {
+export class StyleCollector {
   styles: Map<string, StyleNode | StyleCollector>;
   atRuleName: ?string;
 
--- a/src/sx/src/__tests__/__snapshots__/renderPageWithSX.test.js.snap
+++ b/src/sx/src/__tests__/__snapshots__/renderPageWithSX.test.js.snap
@@ -151,6 +151,40 @@ class="wUqnh _4fo5TC"
 
 `;
 
+exports[`matches expected output: media-queries-goes-last.json 1`] = `
+~~~~~~~~~~ INPUT ~~~~~~~~~~
+{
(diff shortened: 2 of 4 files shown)


tbergq:
This is a specificity issue, maybe we should rather raise the specificity of the css  selectors of media queries

https://codesandbox.io/s/affectionate-kalam-jv2w0?file=/index.html

michalsanger:
So you would render every selector inside media query twice to increase the specificity? Maybe `!important` would do the same job. On the other hand I thought it's a standard to have media queries at the end of a style sheet.

tbergq:
Usage of important should be avoided as far as possible. The approach of repeating the class name is the preferred approach over important also in styled components. I also saw this approach in the new Facebook.com.

tbergq:
https://styled-components.com/docs/advanced#issues-with-specificity

michalsanger:
Let's try that approach #1088

mrtnzlml:
@tbergq Isn't specificity, in this case, being solved by the order as Michal initially did? That sounds like a correct CSS solution that doesn't increase the final CSS size and doesn't use any (IMHO unnecessary) hacks. I don't think we are in a situation where we need to start workarounding things. We just need to render the CSS correctly. 🤔

tbergq:
It is possibly solved by the initial PR as well, but this includes looping the styles X times extra, which to me is unnecessary. You are right, technically, it is not a specificity problem, but it can be solved by increasing the specificity. This is also the way Facebook does it

Screenshot 2020-09-28 at 09 42 17


I do not agree that this is a hack, this is a widely used solution.

PR #1091, opened 2020-09-28
Bazel: offload BUILD files

mrtnzlml:
This commit introduces some basic BUILD files to offload the main Bazel branch. It runs Bazel via Bazelisk on CI and currently executes only two jobs: Buldifier to lint BUILD and WORKSPACE files and `bazel build`.

It should not be used yet so there is no documentation for it yet. I will keep offloading things in a non-breaking fashion in other commits.


WORKSPACE.bazel
```diff
@@ -0,0 +1,76 @@
...
+    sha256 = "b725e6497741d7fc2d55fcc29a276627d10e43fa5d0bb692692890ae30d98d00",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.24.3/rules_go-v0.24.3.tar.gz",
+        "https://github.com/bazelbuild/rules_go/releases/download/v0.24.3/rules_go-v0.24.3.tar.gz",
+    ],
+)
+
+load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
+
+go_rules_dependencies()
```
tbergq, line 54: What are we using go_rules for?

mrtnzlml: Buldifier is written in Go and it requires to be set like this, see: https://github.com/bazelbuild/buildtools/blob/e6efbf6df90bec363c3cbd564b72be6c8a309f14/buildifier/README.md

bazel/macros/build_yarn_workspace.bzl
```diff
@@ -0,0 +1,37 @@
...
+            "src/**/*.test.js",
+        ],
+        data = [
+            ":package.json",
+            ":src",
+            "//:babel.config.js",
+            "//:package.json",
+            "@npm//@adeira/babel-preset-adeira",
+        ],
+    )
```
tbergq, line 29: We are not building flow and es-modules outputs?

tbergq (approved):
Looks like a good start 👍

bazel/macros/build_yarn_workspace.bzl
mrtnzlml: Not yet. But I want to figure out how to do it before considering this to be stable. It should not be part of this build step I think. This build is mainly for other parts of the monorepo to be able to use the built versions of our packages.

tbergq: Sure, makes sense, one way is to build `lib` folder and `es` folder, and have main and module point to the respective folders, but I don't know how to put in the flow files yet 🤕

mrtnzlml: Stay tuned, we are going to figure it out. I also don't know how to do it quite yet.

PR #1098, opened 2020-10-01
Sync Relay example Flow config

mrtnzlml:
This should fix at least one of the CI errors in the shipped version, see: https://github.com/adeira/relay-example/runs/1192058222


tbergq (approved):

tbergq:
The pipeline is green again 👏

PR #1099, opened 2020-10-02
SX: Style hashes inside media queries fix

### LINKED ISSUE #1090: SX: generated classes inside media queries
```js
const styles = sx.create({
  style1: {
    "@media (min-width: 1280px)": {
      color: "red"
    },
    color: "blue"
  },
  style2: {
    color: "red"
  }
});

function Example() {
  return (
    <>
      <div className={styles('style1')}>Red on big screen, blue otherwise</div>
      <div className={styles('style2')}>Always red</div>
    </>
  )
}
```

Does not work as expected because of the generated stylesheet:

```css
@media (min-width: 640px) {
  .wUqnh.wUqnh {
    color: #f00;
  }
}
._4fo5TC {
  color: #00f;
}
.wUqnh {
  color: #f00;
}
```

`_4fo5TC` (blue) is never used. 

Instead of `.wUqnh.wUqnh` we probably should generate different hashes for styles inside media query.

michalsanger:
Same styles inside and outside of media queries don't have the same hash anymore.

Closes #1090

### DIFF
--- a/src/sx/package.json
+++ b/src/sx/package.json
@@ -4,7 +4,7 @@
   "homepage": "https://github.com/adeira/universe/tree/master/src/sx",
   "license": "MIT",
   "private": false,
-  "version": "0.12.0",
+  "version": "0.13.0",
   "main": "index",
   "sideEffects": false,
   "module": false,
--- a/src/sx/src/StyleCollector.js
+++ b/src/sx/src/StyleCollector.js
@@ -10,8 +10,9 @@ function hashStylePair(
   styleName: string,
   styleValue?: string | number = '',
   pseudo?: string = '',
+  atRule?: string | null = '',
 ): string {
-  return hashStyle(`${styleName}${styleValue}${pseudo}`);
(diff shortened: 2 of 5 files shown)


tbergq:
I checked the generated styles of the new Facebook, and it seems they do the same. They generate new style-hashes for media queries, but they also use double classes to bump specificity. I am not sure why they would need the double classes though 🤷 

```css
.maa8sdkg {
  border-right: 1px solid var(--divider);
}
@media (max-width: 899px) {
  .tpef5k2p.tpef5k2p {
    border-right: 1px solid var(--divider);
  }
}
```

So this approach looks good to me

tbergq (approved):

tbergq:
> I am not sure why they would need the double classes though

Actually, this might be needed for the media style to be applied.

michalsanger:
I reverted back the specificity bump because it's still needed:
ezgif-3-c7a055ea1e53

tbergq (approved):

PR #1100, opened 2020-10-03
sx: Add some tests with testing-library

tbergq:
This way we can assert on actually applied style instead of just
checking the class name


michalsanger (approved):
Elegant and readable, I really like it.

mrtnzlml:
That's cool, thanks! :)

PR #1068, opened 2020-09-19
Bazel: add basic build structure

mrtnzlml:
The purpose of this PR is to figure out how to use Bazel correctly in our monorepo and what should be the migration strategy. It is being continuously offloaded to master.

Inspiration:

- https://github.com/tbergq/bazel-monorepo
- https://github.com/jin/awesome-bazel
- https://oasisdigital.com/class/bazel
- https://github.com/GerritCodeReview/gerrit/blob/master/WORKSPACE
- https://github.com/vercel/next.js/pull/22267

### TODOs

- [x] setup Buildifier (https://github.com/bazelbuild/buildtools/blob/master/buildifier/README.md)
- [x] consider Shipit changes (do not ship `*.bazel` files) - https://github.com/adeira/universe/pull/1083
- [x] add Bazel files to `.npmignore` - https://github.com/adeira/universe/pull/1084
- [ ] auto-generate `BUILD.bazel` files (?)
- [ ] update docs (readme)
- [ ] fix CI cache issue
- [ ] run all scripts via Bazel and remove our `@babel/register` hack
- [ ] run Next.js via Bazel (https://github.com/vercel/next.js/search?q=bazel&type=issues)
- [x] use the latest Yarn and Node versions
- [ ] test multiple Node versions (?)
- [ ] get rid of `package.json` files (we need to replace the publishing process)
- [ ] verify testing of related repositories (change in `js` should test everything related?)
- [ ] do something with our `@babel/register` hack

### Possible migration strategy

- [ ] run all the tests via Bazel
- [ ] build and NPM publish all packages via Bazel (https://bazelbuild.github.io/rules_nodejs/Built-ins.html#pkg_npm)
- [ ] lint all packages via Bazel (?)
- [ ] _tktk_

PR #1116, opened 2020-10-06
Bazel: offload more BUILD files

mrtnzlml:
See: https://github.com/adeira/universe/pull/1068

### DIFF
--- a/src/relay-runtime/BUILD.bazel
+++ b/src/relay-runtime/BUILD.bazel
@@ -1,0 +1,14 @@
+load("//:bazel/macros/build_yarn_workspace.bzl", "build_yarn_workspace")
+load("//:bazel/macros/test_yarn_workspace.bzl", "test_yarn_workspace")
+
+build_yarn_workspace(
+    name = "relay-runtime",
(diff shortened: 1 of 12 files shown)


src/relay-runtime/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
...
+
+build_yarn_workspace(
+    name = "relay-runtime",
+    package_name = "@adeira/relay-runtime",
+)
+
+test_yarn_workspace(
+    name = "test",
+    deps = [
+        "@universe//src/js",
```
tbergq, line 12: This makes the source code of adeira/js available, and the test will transpile it?

Would it make sense to use the build target instead?

mrtnzlml: It could be like this eventually, I was thinking about it as well. But now I am trying to keep it as close as possible to what we have now.

tbergq: Sure, we can always improve later 😊

tbergq (approved):

PR #1121, opened 2020-10-08
SX: refactor StyleCollector

mrtnzlml:
The goal of this commit is to cleanup the style collector and prepare it for runtime styles. Specifically, I didn't like that the collector is mingling StyleNode and StyleCollector itself together. So I separated these concerns into individual nodes. These nodes also know how to print themselves so we don't need the CSS renderers.

I also modified the interface of the collector to return style buffer and hash registry instead of mutating the input. This simplified the usage even further.


src/sx/src/__tests__/__snapshots__/renderPageWithSX.test.js.snap
```diff
@@ -23,23 +23,23 @@ exports[`matches expected output: @supports.json 1`] = `
 
 ~~~~~~~~~~ OUTPUT ~~~~~~~~~~
 @supports (display: grid) {
-  ._3GKV06 {
+  ._3GKV06._3GKV06 {
```
mrtnzlml, line 26: Is this correct? Why was this implemented only for `@media` queries? 🤔

tbergq: Yeah, shouldn't be necessary here 🙈

src/sx/src/StyleCollector.js
```diff
@@ -1,94 +1,106 @@
...
+          const nodes = traverse(styleSheetName, maybeValue, new Set([]), maybeName);
+          styleBuffer.add(new StyleCollectorAtNode(maybeName, nodes));
         } else {
-          const isPseudo = property.startsWith(':');
-          iterateEntries(
-            Object.entries(styleValue),
-            isPseudo ? property : pseudo,
-            isPseudo ? className : property,
-          );
+          throw new Error(`Unsupported rule "${maybeName}"`);
```
tbergq, line 74: I am wondering if throwing is a little bit too harsh, this could cause runtime errors, would maybe throwing only when __DEV__ === true be enough?

src/sx/src/StyleCollectorAtNode.js
```diff
@@ -0,0 +1,50 @@
...
+  atRuleName: string;
+  nodes: Set<PrintableNode>;
+
+  constructor(atRuleName: string, nodes: Set<PrintableNode>) {
+    this.atRuleName = atRuleName;
+    this.nodes = nodes;
+  }
+
+  // eslint-disable-next-line no-unused-vars
+  print(config?: PrintConfig): string {
```
tbergq, line 43: Will this config be used later?

src/sx/src/__tests__/__snapshots__/renderPageWithSX.test.js.snap
mrtnzlml: Shouldn't? Cannot there be a case where it's needed just like with media queries?

tbergq: Ah, yes you are right, I was not thinking it through. You could have `display:flex`, but if browser supports `grid`, you want `display:grid`, and it would need higher specificity 👍

tbergq: I am also wondering

```
const styles = sx.create({
  text: {
    color: 'pink',
    ':hover': {color: 'cyan'},
    '@media(min-width:900px)': {color: 'blue'}
  }
})
```
in this case, color pink has specificity (0,0,1,0), and hover and media have (0,0,2,0), so the one defined last in the stylesheet will win, meaning it is not certain that it gets color cyan on hover.

src/sx/src/StyleCollectorAtNode.js
mrtnzlml: It won't, but there is an interface `PrintableNode` so Flow requires it. However, Eslint complains it's unused. It seems like I have to suppress one of them. 🤔

src/sx/src/StyleCollector.js
mrtnzlml: improved

tbergq (approved):
Looks good, nice improvement!

PR #1122, opened 2020-10-08
SX: fix style buffer collecting and resetting

mrtnzlml:
I broke this here: https://github.com/adeira/universe/commit/692f3e1c0c2383663bf39de3eb087df0b2022f43 (Relay example was using the old SX version and I incorrectly thought it works fine)

The style buffer must be initialized only once and reset only when needed (explicitly) because it must remember the styles across the whole application. Moreover, the collected styles must be merged in the style buffer rather than adding them to the set.

Test plan: make sure the relay example still works as expected.

### DIFF
--- a/src/sx/src/__tests__/__snapshots__/renderPageWithSX.test.js.snap
+++ b/src/sx/src/__tests__/__snapshots__/renderPageWithSX.test.js.snap
@@ -186,15 +186,13 @@ exports[`matches expected output: media-queries-multiple.json 1`] = `
   .uxyoC.uxyoC {
     font-size: 1rem;
   }
+  .v2909.v2909 {
+    color: #008000;
+  }
 }
 @media (min-width: 30em) and (max-width: 50em) {
   ._27Xk5F._27Xk5F {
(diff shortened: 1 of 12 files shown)


tbergq (approved):

src/sx/src/__tests__/__snapshots__/renderPageWithSX.test.js.snap
```diff
@@ -186,17 +186,15 @@ exports[`matches expected output: media-queries-multiple.json 1`] = `
...
 @media (min-width: 30em) and (max-width: 50em) {
   ._27Xk5F._27Xk5F {
     color: #00f;
   }
 }
-@media print {
-  .v2909.v2909 {
-    color: #008000;
-  }
-}
```
mrtnzlml, line 199: This is not correctly merged together.

mrtnzlml:
@tbergq Could you please have a look once more? There were more changes to make it work properly.

tbergq (approved):

PR #1129, opened 2020-10-11
SXT Babel plugin: Transform helpers to direct call

michalsanger:
The plugin replace usage of SX Tailwind helpers with direct SX usage. This way the huge SX Tailwind styles can be omitted from a final bundle.


src/babel-plugin-transform-sx-tailwind/src/__tests__/fixtures/sxt-aliased/output.js
```diff
@@ -0,0 +1,49 @@
+// @flow
+import React, { type Element } from 'react';
```
tbergq, line 2: I thought our lint rule would require a space between `@flow` and the first import statement. 

I am not saying you should fix it, I am just curious as to why it is not reporting error here 🤔

src/babel-plugin-transform-sx-tailwind/src/index.js
```diff
@@ -0,0 +1,92 @@
...
+            if (imported.name === 'tailwind') {
+              tailwindCallee = local.name;
+            } else if (imported.name === 'sxt') {
+              sxtCallee = local.name;
+            }
+          });
+
+          const sxImport = template.ast(`import * as sx from '@adeira/sx'`);
+          path.replaceWith(sxImport);
+        }
```
tbergq, line 29: Will it work with `import { create } from '@adeira/sx'` and `import { create as createStyles } from '@adeira/sx'`?

src/babel-plugin-transform-sx-tailwind/src/__tests__/fixtures/sxt-aliased/output.js
michalsanger: I disabled ESLint for these fixtures in `.eslintignore`

src/babel-plugin-transform-sx-tailwind/src/index.js
michalsanger: Good point! Fixed and tested in `fixtures/sx-imported-alias/`

src/babel-plugin-transform-sx-tailwind/src/__tests__/fixtures/sxt-aliased/output.js
tbergq: That would explain it 💡 any reason for ignoring these files?

tbergq: Ok, I see, it is necessary, if not the tests will fail from the eslint formatting.

tbergq (approved):

.flowconfig
```diff
@@ -4,6 +4,7 @@
 <PROJECT_ROOT>/src/docs/build/
 .+/.+\.gzip\.js
 .+/\.next/.+
+<PROJECT_ROOT>/src/babel-plugin-transform-sx-tailwind/src/__tests__/fixtures
```
mrtnzlml, line 7: Would it be possible to make the fixtures valid (real-world examples) to avoid this? 🤔

michalsanger: Actually yes. Not outputs but inputs yes.

PR #1156, opened 2020-10-19
eslint-config: Turn on flowtype/require-readonly-react-props

tbergq:
It seems like it is finally working

### DIFF
--- a/src/eslint-config-adeira/CHANGELOG.md
+++ b/src/eslint-config-adeira/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Remove `__` global, legacy of Nitrolib from Kiwi.com, see: https://kiwicom.github.io/nitrolib/services.html#intl
+- Set `flowtype/require-readonly-react-props` to warning in normal mode and error in strict mode
 
 # 3.2.0
 
--- a/src/eslint-config-adeira/__tests__/__snapshots__/index.test.js.snap
+++ b/src/eslint-config-adeira/__tests__/__snapshots__/index.test.js.snap
@@ -151,7 +151,7 @@ Object {
     "flowtype/require-indexer-name": 0,
     "flowtype/require-inexact-type": 2,
     "flowtype/require-parameter-type": 0,
-    "flowtype/require-readonly-react-props": 0,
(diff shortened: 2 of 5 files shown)


src/eslint-config-adeira/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Remove `__` global, legacy of Nitrolib from Kiwi.com, see: https://kiwicom.github.io/nitrolib/services.html#intl
+- Set `flowtype/require-readonly-react-props` to NEXT_VERSION_ERROR
```
mrtnzlml, line 4: `NEXT_VERSION_ERROR` is quite an internal thingy. I usually write something like "warnings or errors in strict mode" for the public. Would you please consider it?

tbergq: Yes, you are right, I will update it

tbergq: 🚀

mrtnzlml (approved):
I am wondering: is it working now OR we are not using any of the cases mentioned here? https://github.com/gajus/eslint-plugin-flowtype/pull/400#issuecomment-492828739

tbergq:
> I am wondering: is it working now OR we are not using any of the cases mentioned here? [gajus/eslint-plugin-flowtype#400 (comment)](https://github.com/gajus/eslint-plugin-flowtype/pull/400#issuecomment-492828739)

I tested with `props: $FlowFixMe` and it didn't produce any error. I didn't specifically try for the rest though. 

But they were for relay things, right? And we should have gotten errors for that in `example-relay`?

mrtnzlml:
@tbergq There is a chance that when I was trying it, we had much more React code in Universe. Perhaps `src/__flowtests__` would be a good place to verify these things?

But not that big deal - 🚀

tbergq:
> @tbergq There is a chance that when I was trying it, we had much more React code in Universe. Perhaps `src/__flowtests__` would be a good place to verify these things?
> 
> But not that big deal - 🚀

Yeah, possibly there was more react code here. I tested in in `kiwicom/account` and no unexpected errors there either

PR #1164, opened 2020-10-20
SX: disallow nested pseudo classes/elements

mrtnzlml:
This change disallows cases like:

```text
{
  ':hover': {
    'color': 'red',
    ':active': {
      color: 'blue'
    }
  }
}
```

Instead, you can write this:

```text
{
  ':hover': {
    color: 'red',
  },
  ':active': {
    color: 'blue'
  }
}
```


src/sx/src/StyleCollector.js
```diff
@@ -61,6 +61,9 @@ class StyleCollector {
           }
         } else if (maybeName.startsWith(':')) {
           // pseudo type (:hover, ::after, ...)
+          for (const key of Object.keys(maybeValue)) {
+            invariant(key.startsWith(':') === false, 'Nested pseudo classes are not allowed.');
```
tbergq, line 65: what about`::after:hover`?

mrtnzlml: Do you mean when you write it as one selector? I actually didn't try it but I don't see a problem with that. It just cannot be nested. Will add some tests for it as well.

tbergq: Aha, of course, it is not nested in this case 🤦

tbergq (approved):

src/sx/src/StyleCollector.js
mrtnzlml: I added a test to cover this scenario:

```text
exports[`matches expected output: pseudo-classes-nested-2.js 1`] = `
~~~~~~~~~~ INPUT ~~~~~~~~~~
{
  "test": {
    ":hover::after": {
      "content": "this is OK"
    }
  }
}
~~~~~~~~~~ OUTPUT ~~~~~~~~~~
._2BDpfz:hover::after {
  content: "this is OK";
}

~~~~~~~~~~ USAGE ~~~~~~~~~~

className={styles('test')}
  ↓ ↓ ↓
class="_2BDpfz"

`;
```

tbergq (approved):

PR #1178, opened 2020-10-21
Upgrade react to 17.0.0 and upgrade next to 9.5.5

tbergq:


package.json
```diff
@@ -37,6 +37,8 @@
     "typecheck": "yarn run adeira-flow-bin"
   },
   "resolutions": {
-    "graphql": "^15.0.0"
+    "graphql": "^15.0.0",
+    "react": "^17.0.0",
+    "react-dom": "^17.0.0"
```
tbergq, line 42: Since storybook drags in react 16.8.3 and makes `example-relay` fail

tbergq: Storybook still works with react 17.0.0

mrtnzlml: I see there are some issues with Storybook and React 17, for example: https://github.com/storybookjs/storybook/issues/12849

Would it make sense to wait for the Storybook to catch up?

tbergq: We could, but our storybook still works. If you think it is better to wait, we can wait.

mrtnzlml: I'd wait a little bit longer. We could maybe offload the `next` upgrade in a different PR? 🤔

michalsanger: Upgrade to Next 10? :-)

PR #1213, opened 2020-10-28
sx: Remove auto adding of quotes in content

tbergq:
- Render generated css with dangerouslySetInnerHtml

It is necessary to specify the quoutes of the content, since
usage with a counter will not work if we automatically add the qoute.


mrtnzlml (approved):
👍

PR #1214, opened 2020-10-28
SXT Website: More examples

michalsanger:


src/sx-tailwind-website/src/TailwindCss/Buttons.js
```diff
@@ -0,0 +1,57 @@
+// @flow
+
+import * as React from 'react';
+// eslint-disable-next-line no-unused-vars
+import * as sx from '@adeira/sx';
```
tbergq, line 5: Why do you need to import this?

mrtnzlml (approved):
I had and still have the same question as @tbergq but otherwise 👍

src/sx-tailwind-website/src/TailwindCss/Buttons.js
michalsanger: It's a workaround until I find out what is going on during the Next.js build. It was failing with `ReferenceError: sx is not defined`. There are some optimisation happening, source code of imported components is pasted into code that was importing them and for some reason it does not play well with that Tailwind to SX Babel transformation.

Do you have any idea what could be the issue? cc @mrtnzlml

michalsanger: Fixed in https://github.com/adeira/universe/pull/1220

PR #1215, opened 2020-10-29
SX: hash already transformed values

mrtnzlml:
Previously, we were hashing raw style names and values which could result in duplicated CSS classes. We should hash already transformed style names and values instead. This commit fixes it.


mrtnzlml:
This, unfortunately, results in many changes in hashes so please pay attention to the details.

tbergq (approved):

PR #1217, opened 2020-10-29
SX: add `font-size` to the CSS reset

mrtnzlml:
SX expects font size to be exactly 16px (otherwise px->rem conversion won't work correctly). This means it should be included by default. Which makes me wonder: should we include the CSS reset by default and make it opt-out?

### DIFF
--- a/src/sx/README.md
+++ b/src/sx/README.md
@@ -184,7 +184,9 @@ import Document from 'next/document';
 
 export default class MyDocument extends Document {
   static getInitialProps(ctx: DocumentContext) {
-    return sx.renderPageWithSX(ctx.renderPage, { includeReset: true });
+    return sx.renderPageWithSX(ctx.renderPage, {
+      includeReset: true, // <<<
+    });
   }
 
   // `render` is not needed to change
@@ -194,9 +196,12 @@ export default class MyDocument extends Document {
 This will add the following css to your stylesheet:
 
 ```css
+html,
 body {
+  font-size: 16px;
   box-sizing: border-box;
 }
+
 *,
 *::after,
 *::before {
@@ -205,6 +210,8 @@ body {
   box-sizing: inherit;
 }
 ```
+
+You might wonder why does it include the `font-size: 16px`? It's because SX automatically converts numbers to REM units and it expects `16px` to be the document base.
 
 ## Architecture
 
--- a/src/sx/src/cssReset.js
+++ b/src/sx/src/cssReset.js
@@ -1,0 +1,25 @@
+// @flow
+
+import csstree from 'css-tree';
+
+function css(strings): string {
+  const css = strings.join('');
+  return csstree.generate(csstree.parse(css));
+}
+
+// TODO: consider including https://github.com/necolas/normalize.css
+export default (css`
+  html,
+  body {
+    font-size: 16px; /* SX expects 16px to be the default (see "transformValue") */
+    box-sizing: border-box;
+  }
+
+  *,
+  *::after,
+  *::before {
+    margin: 0;
+    padding: 0;
+    box-sizing: inherit;
+  }
+`: string);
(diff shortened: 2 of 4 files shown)


src/sx/README.md
```diff
@@ -194,9 +196,12 @@ export default class MyDocument extends Document {
 This will add the following css to your stylesheet:
 
 ```css
+html,
 body {
+  font-size: 16px;
```
tbergq, line 201: It should be sufficient to have it only on the HTML tag

src/sx/src/cssReset.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+
+import csstree from 'css-tree';
+
+function css(strings): string {
+  const css = strings.join('');
+  return csstree.generate(csstree.parse(css));
+}
+
+// TODO: consider including https://github.com/necolas/normalize.css
```
tbergq, line 10: I am not sure we should mix reset and normalize, since they are 2 different ideas, but we could add an option for choosing between normalize and reset.

tbergq:
> SX expects font size to be exactly 16px

Also, I am wondering if we should just drop this. Now we only calculate font-sizes, right? I might want to use rems for padding and margin and what not. By setting the font-size to `62.5%` on the HTML tag, that makes 1rem = 10px and it is easy to control for the dev. Now I cannot do that, since it messes with the font sizes 😬

src/sx/src/cssReset.js
michalsanger: I'd add normalisation, I'm already using it

tbergq: Yeah, that is why we could allow the dev to choose between CSS reset and normalize.

mrtnzlml: @tbergq I would not like to make it more complex with additional configuration options and choices. It would be nice if SX provided a good to go default and that's it (so you don't have to add these resets/normalizers on your own). At least for now before finishing the basics.

mrtnzlml:
> > SX expects font size to be exactly 16px
> 
> Also, I am wondering if we should just drop this. Now we only calculate font-sizes, right? I might want to use rems for padding and margin and what not. By setting the font-size to `62.5%` on the HTML tag, that makes 1rem = 10px and it is easy to control for the dev. Now I cannot do that, since it messes with the font sizes 😬

@tbergq The main motivation for this was https://engineering.fb.com/web/facebook-com-accessibility/ specifically "Scalable font sizes" section. How would you imagine it to be if we had to change it?

tbergq:
> > > SX expects font size to be exactly 16px
> > 
> > 
> > Also, I am wondering if we should just drop this. Now we only calculate font-sizes, right? I might want to use rems for padding and margin and what not. By setting the font-size to `62.5%` on the HTML tag, that makes 1rem = 10px and it is easy to control for the dev. Now I cannot do that, since it messes with the font sizes 😬
> 
> @tbergq The main motivation for this was https://engineering.fb.com/web/facebook-com-accessibility/ specifically "Scalable font sizes" section. How would you imagine it to be if we had to change it?

Yeah, I am all for Scalable font sizes, and it is important for a lot of users. I am just saying there are other ways of achieving it, like the way mentioned above. Also, we can keep this default,  but it would be nice to apply it to other style rules like margin and padding as well. But that can be improved later.

src/sx/README.md
tbergq: Actually, I guess it should be 

```css
html {
  font-size: 16px;
}
body {
  font-size: 1rem;
}
```

src/sx/src/cssReset.js
tbergq: >  I would not like to make it more complex 

Yeah, I am fine with that 👍 

> It would be nice if SX provided a good to go default

Exactly, but I don't think we should include reset and normalize at once.

src/sx/README.md
tbergq: Actually, even setting `font-size: 16px` on the HTML prevents scaling font-sizes. It should be set to `100%` or not set at all.

mrtnzlml:
https://github.com/adeira/universe/pull/1261

PR #1218, opened 2020-10-29
relay-example: Fix ordered list styling

tbergq:


mrtnzlml:
@tbergq This breaks bi-directional pagination example when you go to a second page (currently the second is from 21-40 but with this change it starts from zero again).

tbergq:
> @tbergq This breaks bi-directional pagination example when you go to a second page (currently the second is from 21-40 but with this change it starts from zero again).

Yeah, you are right 🙈 I only tested with the other two! 

I will see if I can fix it.

tbergq:
@mrtnzlml Please check again, it should be correct now.

src/example-relay/src/Homepage/locations/LocationsList.js
```diff
@@ -10,10 +10,12 @@ type Props = {|
...
 |};
 
-export default function LocationList({ children, ...rest }: Props): React.Node {
+export default function LocationList({ children, start = 1 }: Props): React.Node {
   return (
-    <ol {...rest} className={styles('ol')}>
-      {children}
+    <ol start={start} className={styles('ol')}>
+      {React.Children.map(children, (child, i) =>
+        React.cloneElement(child, { liClass: styles('li'), dataCount: start + i }),
```
mrtnzlml, line 17: Why is the `li` class being styled here (and passed down) rather than near the actual `li` element? 🤔

tbergq: Yeah, with the initial solution I think it made more sense because I had to put the counter-reset on the `ol` and use it also on the li, so it would be less `wtf`, where does this come from. 

I think now with this solution, I can move the styling to the `li` instead

tbergq: 🚀

mrtnzlml (approved):

PR #1220, opened 2020-10-29
SXT Babel: Enable usage within monorepo

michalsanger:
This enables usage of the `babel-plugin-transform-sx-tailwind` within monorepo. Local deps are just symlinked which can be an issue if a transpilation must be done.

It also adds styles collector reset when a new file is processed, this causes crashes during build.

### DIFF
--- a/src/babel-plugin-transform-sx-tailwind/src/index.js
+++ b/src/babel-plugin-transform-sx-tailwind/src/index.js
@@ -1,12 +1,17 @@
 // @flow
 
-import template from '@babel/template';
-import * as t from '@babel/types';
-import murmurHash from '@adeira/murmur-hash';
-import { tailwindStyles } from '@adeira/sx-tailwind';
+require('@babel/register')({
+  ignore: [/node_modules\/(?!@adeira)/],
+  rootMode: 'upward',
+});
 
-module.exports = function sxTailwindBabelPlugin(): any {
-  const stylesCollector = [];
+const template = require('@babel/template').default;
+const t = require('@babel/types');
+const murmurHash = require('@adeira/murmur-hash').default;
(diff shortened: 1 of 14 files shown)


src/sx-tailwind-website/package.json
```diff
@@ -16,16 +16,15 @@
...
-    "serve": "^11.3.2"
+    "flow-bin": "^0.136.0"
   },
   "scripts": {
     "dev": "next",
     "build": "next build",
     "preexport": "yarn run build",
     "export": "next export",
-    "start": "serve out"
+    "start": "next start"
```
mrtnzlml, line 28: Do you still need the `export` and `preexport`? I checked the Vercel configuration and maybe we could change the preset back from `Other` to default `Next.js`?

src/babel-plugin-transform-sx-tailwind/src/index.js
```diff
@@ -62,6 +67,7 @@ module.exports = function sxTailwindBabelPlugin(): any {
       },
       Program: {
         enter(path, state) {
+          stylesCollector = [];
```
mrtnzlml, line 70: 👍 I think this was the main culprit. Is there a simple way how to test it?

mrtnzlml (approved):

src/sx-tailwind-website/package.json
michalsanger: Yep, I'll try to switch to default Nextjs settings.

src/babel-plugin-transform-sx-tailwind/src/index.js
michalsanger: Exactly, it was collecting styles incrementally from all files. Unfortunately no idea how to test it :-(

PR #1234, opened 2020-10-31
Adeira.dev: freshen-up the design

mrtnzlml:
I also removed unnecessary/unused files and added 2 new links to the menu promoting our other projects (SX Tailwind website and Relay example).

Figma: https://www.figma.com/file/bAVVTRg9w2vDJ1Hph82hky/Adeira


mrtnzlml:
After this task, I'd like to update the main README file as well (since it's a bit connected). 🚀

michalsanger (approved):

src/adeira.dev/static/css/custom.css
```diff
@@ -1,76 +1,29 @@
...
-}
-
-.iconContainer {
-  margin-bottom: 30px;
-  margin-right: 30px;
-  max-width: 200px;
-  float: left;
+  background-color: #55e0c6;
+  padding: 40px;
+  color: #000;
```
tbergq, line 17: Should we have used the same black as the rest of the project here?

`var(--ifm-font-color-base)`

And, it doesn't change to white when I switch to dark mode, or is that intentional?

mrtnzlml: It's intentional not to change to color. The background is too bright for doing so - it looked weird.

tbergq: > The background is too bright for doing so - it looked weird.

👍

PR #1239, opened 2020-10-31
SX: expand shorthand properties

mrtnzlml:
Purpose of this change is to expand shorthand CSS properties which could cause conflics when defined together. Imagine the following example:

```
const styles = sx.create({
  primary: { marginTop: '10px' },
  button: { margin: 0 },
});
```

It could generate the following CSS:

```
.c0 { margin-top: 10px }
.c1 { margin: 0px }
```

That would be a tricky situation because CSS specificity depends on the CSS definition order and therefore the top margin would be always overwritten when used together. Instead, we expand the shorthand properties into something like this:

```
.c0 { margin-top: 10px }   << primary
.c1 { margin-top: 0px }    << button
.c2 { margin-right: 0px }
.c3 { margin-bottom: 0px }
.c4 { margin-left: 0px }
```

This way SX is not dependent on the CSS rules insertion order but rather on the markup.

Inspired by this article: https://sebastienlorber.com/atomic-css-in-js (and React Native Web)

### DIFF
--- a/src/sx/src/expandShorthandProperties.js
+++ b/src/sx/src/expandShorthandProperties.js
@@ -1,0 +1,140 @@
+// @flow
+
+import StyleCollectorNode from './StyleCollectorNode';
+import { isColor } from './colorNormalizer';
+
+/**
+ * Purpose of this function is to expand shorthand CSS properties which could cause conflics
+ * when defined together. Imagine the following example:
+ *
+ * ```
+ * const styles = sx.create({
+ *   primary: { marginTop: '10px' },
+ *   button: { margin: 0 },
+ * });
+ * ```
+ *
+ * It could generate the following CSS:
+ *
+ * ```
+ * .c0 { margin-top: 10px }
+ * .c1 { margin: 0px }
+ * ```
+ *
+ * That would be a tricky situation because CSS specificity depends on the CSS definition order
+ * and therefore the top margin would be always overwritten when used together. Instead, we
+ * expand the shorthand properties into something like this:
+ *
+ * ```
+ * .c0 { margin-top: 10px }   << primary
+ * .c1 { margin-top: 0px }    << button
+ * .c2 { margin-right: 0px }
+ * .c3 { margin-bottom: 0px }
+ * .c4 { margin-left: 0px }
+ * ```
+ *
+ * This way SX is not dependent on the CSS rules insertion order but rather on the markup.
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties
+ */
+export default function expandShorthandProperties(
+  propertyName: string,
+  propertyValue: any,
+  hashSeed: string = '',
+): $ReadOnlyArray<StyleCollectorNode> {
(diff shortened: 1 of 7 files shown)


src/sx/src/expandShorthandProperties.js
```diff
@@ -0,0 +1,122 @@
...
+  hashSeed: string = '',
+): $ReadOnlyArray<StyleCollectorNode> {
+  // https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties#Background_Properties
+  if (propertyName === 'background') {
+    // Note: we ignore more complex syntaxes at this moment
+    // see: https://developer.mozilla.org/en-US/docs/Web/CSS/background#Formal_syntax
+    if (isColor(propertyValue)) {
+      return [new StyleCollectorNode('backgroundColor', propertyValue, hashSeed)];
+    } else if (propertyValue === 'none') {
+      return [new StyleCollectorNode('backgroundImage', propertyValue, hashSeed)];
```
michalsanger, line 52: Shouldn’t it also set `background-color` to transparent?

mrtnzlml: @michalsanger I was writing it based on this [formal syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/background#Formal_syntax) and it seems like value `none` applies exclusively to `background-image`. This SO question supports my thinking: https://stackoverflow.com/q/20784292/3135248

michalsanger: On the SO is also this answer which makes sense to me:
> So... when you set: background:none;
> You are saying that background-image: none; and all the others to the initial state (as they are not being declared).

mrtnzlml: I understood from the answer that only the bg-image prop is explicitly set and the rest of them is set implicitly by the browser (default values like when you don't write anything). Are you suggesting we should set _all_ the properties explicitly? I am not sure that's correct. 🤔

I was creating this PR to solve one issue with runtime styles and this `background:none` change solved the issue.

I am summoning @tbergq. Do you have any more knowledge regarding this?

michalsanger: In this example the background would still be blue no? Isn't confusing?
```js
const styles = sx.create({
  bgBlue: {
    background: 'blue',
  },
  bgNone: {
    background: 'none',
  },
});

<div className={styles('bgBlue', 'bgNone')}>Lorem lipsum</div>
```

mrtnzlml: I see. That would be confusing. So when writing normal CSS, the browser is doing the implicit work. But in SX expansion we are the ones who have to do it, right?

michalsanger: Yes, I'd say so

mrtnzlml: @michalsanger Please, have a look again. I changed the expansion logic so it expands into all bg properties.

michalsanger (approved):

PR #1250, opened 2020-11-02
Use new JSX transform

tbergq:
https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -35,6 +35,7 @@ module.exports = function (api /*: ApiType */) /*: BabelConfig */ {
         '@adeira/babel-preset-adeira',
         {
           target: api.caller(isWebpack) ? 'js-esm' : 'js',
+          reactRuntime: 'automatic',
         },
       ],
     ],
--- a/src/eslint-config-adeira/ourRules.js
+++ b/src/eslint-config-adeira/ourRules.js
@@ -429,7 +429,7 @@ module.exports = ({
   'react/prefer-read-only-props': OFF,
   'react/prefer-stateless-function': OFF,
   'react/prop-types': OFF, // we use Flow instead,
-  'react/react-in-jsx-scope': WARN,
+  'react/react-in-jsx-scope': OFF,
   'react/require-default-props': OFF,
   'react/require-optimization': OFF,
   'react/require-render-return': ERROR,
--- a/src/relay/CHANGELOG.md
+++ b/src/relay/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Change to use the new [JSX transforms](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)
+
 # 2.1.0
 
 - Added new Flow utils `FragmentContainerType`, `RefetchContainerType` and `PaginationContainerType` which will help you to migrate to the new `types-first` Flow architecture (https://medium.com/flow-type/types-first-a-scalable-new-architecture-for-flow-3d8c7ba1d4eb). Example:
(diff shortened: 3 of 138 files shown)


src/eslint-config-adeira/ourRules.js
```diff
@@ -429,7 +429,7 @@ module.exports = ({
   'react/prefer-read-only-props': OFF,
   'react/prefer-stateless-function': OFF,
   'react/prop-types': OFF, // we use Flow instead,
-  'react/react-in-jsx-scope': WARN,
+  'react/react-in-jsx-scope': OFF,
```
tbergq, line 432: Possibly we should leave this and just update the rule directly in `.eslintrc.js`, and change it here with the next major version

mrtnzlml: File `src/eslint-config-adeira/getCommonConfig.js` has `detectReactVersion` config. Would that be helpful?

tbergq: But using the new `JSX` transform is still optional, so I don't think so.

babel.config.js
```diff
@@ -35,6 +35,7 @@ module.exports = function (api /*: ApiType */) /*: BabelConfig */ {
         '@adeira/babel-preset-adeira',
         {
           target: api.caller(isWebpack) ? 'js-esm' : 'js',
+          reactRuntime: 'automatic',
```
mrtnzlml, line 38: We should probably swich it to `automatic` by default after this PR. Followup issue? :)

tbergq: #1253

src/relay/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Change to use the new [JSX transforms](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)
```
mrtnzlml, line 3: Are there any implications for the user-land code? Does it require to use the new JSX transform as well?

tbergq: No, it does not. But it needs to be using a version compatible with the new `JSX` transform. So 16.14.0 or greater

tbergq: The transpilde code looks like 

```jsx
import {jsx as _jsx} from 'react/jsx-runtime';

function App() {
  return _jsx('h1', { children: 'Hello world' });
}
```

So as long as that import path exists in react, the consumer is fine.

src/sx-tailwind-website/__github__/babel.config.js
```diff
@@ -10,8 +10,8 @@ type ApiType = {|
 |};
 
 type BabelConfig = {|
-  +presets: $ReadOnlyArray<string>,
-  +plugins: $ReadOnlyArray<string>,
+  +presets: $ReadOnlyArray<string | [string, { ... }]>,
+  +plugins: $ReadOnlyArray<string >,
```
mrtnzlml, line 14:
```suggestion
  +plugins: $ReadOnlyArray<string>,
```

src/eslint-config-adeira/ourRules.js
tbergq: We are just flipping it from `WARN` -> `OFF`, so I think this approach is fine.

src/sx-tailwind-website/__github__/babel.config.js
tbergq: Done

mrtnzlml (approved):

PR #1255, opened 2020-11-02
SX: expand (simple) shorthand border property

mrtnzlml:


mrtnzlml:
@michalsanger Is the Vercel comment going to be under every PR even though it's unrelated (the deployable paths didn't change)? Is there a way how to improve it?

michalsanger:
@mrtnzlml looks like yes, the only improvement available is to skip the build when the changes are unrelated.

michalsanger:
OK, can be turned off - https://vercel.com/knowledge/how-to-prevent-vercel-github-comments

But probably for every PR :-(

michalsanger (approved):

PR #1261, opened 2020-11-03
SX: remove CSS reset

mrtnzlml:
We couldn't decide whether to use CSS reset or normalization and concluded that we should leave this decision for the user.

### DIFF
--- a/src/example-relay/public/reset.css
+++ b/src/example-relay/public/reset.css
@@ -1,0 +1,10 @@
+body {
+  box-sizing: border-box;
+}
+*,
+*::after,
+*::before {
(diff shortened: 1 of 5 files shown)


tbergq (commented):
Could you add the reset we had in sx to `relay-example`? Since the styling there now expects this reset

mrtnzlml:
@tbergq done

mrtnzlml:
Another reason why to remove it: it's only SSR and doesn't support runtime styles (in case someone doesn't care about SSR).

tbergq (approved):

PR #1271, opened 2020-11-04
SX: remove 'px' -> 'rem' conversion of font-size

mrtnzlml:
As suggested by Trond: it would be better to leave this up to the developer to choose whatever strategy they want (to allow 62.5% approach for example). Moreover, 16px works only when user doesn't scale the website in a browser settings which is not always the case.


tbergq (approved):
We should probably update relay-example font-sizes when we release the next version of sx

PR #1277, opened 2020-11-05
babel-preset: Set reactRuntime automatic as default

### LINKED ISSUE #1253: babel-preset: Set reactRuntime to automatic by default
TODO:

- [x] land https://github.com/adeira/universe/pull/1250 first

tbergq:
closes #1253

Release @adeira/babel-preset-adeira 2.0.0

### DIFF
--- a/src/monorepo-npm-publisher/CHANGELOG.md
+++ b/src/monorepo-npm-publisher/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Upgrade `@adeira/babel-preset-adeira` to 2.0.0. This has some implications on the transpiled content released to npm from this package. We should not release a new version of this package before we figure out a way to let the user decide which kind of react-transform he/she wants to use.
+
 # 1.0.0
 
 - Updated dependencies
--- a/src/monorepo-npm-publisher/CHANGELOG.md
+++ b/src/monorepo-npm-publisher/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
-- Upgrade `@adeira/babel-preset-adeira` to 2.0.0. This has some implications on the transpiled content released to npm from this package. We should not release a new version of this package before we figure out a way to let the user decide which kind of react-transform he/she wants to use.
+- Upgrade `@adeira/babel-preset-adeira` to 2.0.0.
+- Breaking Change: Add reactRuntime option. You can set it to `automatic` to use the new [JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html), or set it to `classic` to use the classical version. Note that `automatic` is the default.
 
 # 1.0.0
 
--- a/src/monorepo-npm-publisher/package.json
+++ b/src/monorepo-npm-publisher/package.json
@@ -8,7 +8,7 @@
   "main": "src/index",
   "sideEffects": false,
   "dependencies": {
-    "@adeira/babel-preset-adeira": "^1.2.0",
+    "@adeira/babel-preset-adeira": "^2.0.0",
     "@adeira/monorepo-utils": "^0.9.0",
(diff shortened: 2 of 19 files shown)


src/monorepo-npm-publisher/package.json
```diff
@@ -8,7 +8,7 @@
   "main": "src/index",
   "sideEffects": false,
   "dependencies": {
-    "@adeira/babel-preset-adeira": "^1.2.0",
+    "@adeira/babel-preset-adeira": "^2.0.0",
```
mrtnzlml, line 11: I was thinking: could this have some undesired implications on our released NPM packages? Like different React version requirements for example? Should we do it more carefully in such a case?

tbergq: I don't think so, or maybe there is something I am not seeing. Would my mind specify exactly what you fear might happen?

tbergq: Ok, I see what you mean now, and we already created this problem for ourselves, since I forgot to set `reactRuntime: 'automatic'` in `monorepo-npm-publisher`. 

We definitely need to find some solution here before we release a new version of this package.

However, this MR is not releasing this package.

tbergq: I made a note of it in the changelog, under the unreleased section.

yarn.lock
```diff
@@ -9,6 +9,21 @@
   dependencies:
     "@babel/runtime" "^7.11.2"
 
+"@adeira/sx@^0.15.0":
```
mrtnzlml, line 12: Some package doesn't use the latest `@adeira/sx` version? This looks wrong. 🤔

src/monorepo-npm-publisher/package.json
mrtnzlml: @tbergq So from now on the NPM publisher will release React code which is only compatible with the new JSX transform, do I understand it correctly? So every package will require peer React dependency at least 16.4, right?

tbergq: > So from now on the NPM publisher will release React code which is only compatible with the new JSX transform, do I understand it correctly?

Yes, but I think we could add an option to which runtime to target. 

> So every package will require peer React dependency at least 16.4, right?

Well, `16.14.0`

yarn.lock
tbergq: Yeah, this came when rebasing master into this branch. It should be fixed now.

src/monorepo-npm-publisher/package.json
tbergq: Ok, I added the option to configure `reactRuntime` for `monorepo-npm-publisher`.

src/monorepo-npm-publisher/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+- Upgrade `@adeira/babel-preset-adeira` to 2.0.0.
+- Add reactRuntime option. You can set it to `automatic` to use the new [JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html), or set it to `classic` to use the classical version. Note that `automatic` is the default.
```
mrtnzlml, line 4: This is a breaking change, right?

mrtnzlml (approved):

src/monorepo-npm-publisher/CHANGELOG.md
tbergq: Yes, that is correct. The next release should be a major version. Should I mark it here with BC?

mrtnzlml: Yes please, just to be sure we release it correctly.

tbergq: 🚀

PR #1279, opened 2020-11-05
SX Tailwind: Support transitions

michalsanger:
This also adds support for CSS variables into SX.


src/sx/package.json
```diff
@@ -4,7 +4,7 @@
   "homepage": "https://github.com/adeira/universe/tree/master/src/sx",
   "license": "MIT",
   "private": false,
-  "version": "0.15.1",
+  "version": "0.15.2",
```
tbergq, line 7: I think the removal of the CSS reset will be released when this is bumped, so better make it `0.16.0`

src/sx/src/transformStyleName.js
```diff
@@ -3,5 +3,10 @@
 import * as changeCase from 'change-case';
 
 export default function transformStyleName(styleName: string): string {
+  if (styleName.startsWith('--')) {
+    // CSS variable
+    return styleName;
```
tbergq, line 8: It already works with CSS variables, what makes this change necessary?

michalsanger: See the `src/sx/src/__tests__/transformStyleName.test.js`. It was transforming the variable name.

tbergq: Aha, I see, this is for defining the variable, not using it 👍

src/sx/package.json
michalsanger: OK, done

tbergq (approved):

src/sx/package.json
mrtnzlml: In general, I'd say releasing patch versions for 0.x is not necessary - feel free to always bump the minor. Semver understands these versions a bit differently and more aggressively:

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

Also in FAQ:

> How should I deal with revisions in the 0.y.z initial development phase?
>
> The simplest thing to do is start your initial development release at 0.1.0 and then increment the minor version for each subsequent release.

But up to you, ofc. https://semver.org/

PR #1281, opened 2020-11-06
Release @adeira/sx 0.17.0

tbergq:
- Set reactRuntime: 'automatic' in npm-monorepo-publisher

### DIFF
--- a/src/monorepo-npm-publisher/src/transformFileVariants.js
+++ b/src/monorepo-npm-publisher/src/transformFileVariants.js
@@ -28,6 +28,7 @@ export default function transformFileVariants(
           '@adeira/babel-preset-adeira',
           {
             target,
+            reactRuntime: 'automatic',
           },
         ],
       ],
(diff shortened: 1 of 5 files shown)


michalsanger (approved):

src/monorepo-npm-publisher/src/transformFileVariants.js
```diff
@@ -28,6 +28,7 @@ export default function transformFileVariants(
           '@adeira/babel-preset-adeira',
           {
             target,
+            reactRuntime: 'automatic',
```
mrtnzlml, line 31: @tbergq Could you please elaborate on what was this issue here? How is SX or SXTailwind related to React? 🤔 Just trying to understand what went wrong exactly.

tbergq: Yes, the `src/sx/src/renderPageWithSX.js` has 

```jsx
styles: [
      // We need to render this as html, else things like `content: "ok"` will be rendered as `content: &quot;ok&quot;`, and it is not valid css
      <style
        key="adeira-sx"
        data-adeira-sx={true}
        dangerouslySetInnerHTML={{ __html: StyleCollector.print() }}
      />,
    ],
```

When using classical react runtime, this gets transpiled into `React.createElement`, however, this file does not contain a react import, and therefore it broke with the `classical` runtime.

PR #1294, opened 2020-11-10
sx: Add keyframes support

tbergq:
- release `@adeira/sx@0.19.0`

### DIFF
--- a/src/sx/README.md
+++ b/src/sx/README.md
@@ -5,6 +5,7 @@ In conventional applications, CSS rules are duplicated throughout the stylesheet
   - [Multiple stylesheets precedence](#multiple-stylesheets-precedence)
   - [Pseudo CSS classes and elements](#pseudo-css-classes-and-elements)
   - [`@media` and `@supports`](#media-and-supports)
+  - [Keyframes](#keyframes)
   - [Precise Flow types](#precise-flow-types)
 - [Production usage considerations](#production-usage-considerations)
 - [Server-side rendering](#server-side-rendering)
@@ -225,6 +226,41 @@ const styles = sx.create({
     },
   },
 });
+```
+
+### Keyframes
+
+SX also has support for keyframes, it exports a function that generates an animation name from the input you give it. You can use it like this:
+
+```jsx
+export function AnimatedComponent() {
+  return <div className={styles('text')}>text</div>;
+}
+
+const fadeIn = sx.keyframes({
+  '0%': {
+    opacity: 0,
+  },
+  '50%, 55%': {
+    opacity: 0.3,
+  },
+  '100%': {
+    opacity: 1,
+  },
+});
+
+const styles = sx.create({
+  text: {
+    animationName: fadeIn,
+    animationDuration: '2s',
+  },
+});
+```
+
+It also supports `from` and `to` for simpler animations.
+
+```js
+const simple = sx.keyframes({ from: { opacity: 0 }, to: { opacity: 1 } });
 ```
 
 ### Precise Flow types
--- a/src/sx/src/create.js
+++ b/src/sx/src/create.js
@@ -15,15 +15,8 @@ type MediaQueries = {|
   +[string]: MediaQueries, // media queries can be recursively nested
 |};
 
-// https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes
-type KeyFrames = {|
-  +from: AllCSSPropertyTypes,
-  +to: AllCSSPropertyTypes,
-  +[number]: AllCSSPropertyTypes, // percentages
-|};
-
 // https://developer.mozilla.org/en-US/docs/Web/CSS/At-rule
-type AtRules = MediaQueries | KeyFrames;
+type AtRules = MediaQueries;
 
 type AllCSSProperties = {|
   ...AllCSSPropertyTypes,
--- a/src/sx/src/injectRuntimeStyles.js
+++ b/src/sx/src/injectRuntimeStyles.js
@@ -13,8 +13,7 @@ import type { StyleBufferType } from './StyleCollector';
 opaque type StyleElementType = HTMLStyleElement | null;
 let styleAdeiraSXTag: StyleElementType = null;
 
-// https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model
-export default function injectRuntimeStyles(styleBuffer: StyleBufferType) {
+const getStyleTag = (): CSSStyleSheet => {
   if (styleAdeiraSXTag === null) {
     styleAdeiraSXTag = ((document.querySelector('style[data-adeira-sx]'): any): StyleElementType);
     if (styleAdeiraSXTag === null) {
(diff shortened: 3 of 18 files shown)


src/example-relay/src/components/Navbar.js
```diff
@@ -1,5 +1,7 @@
 // @flow
 
+/* eslint-disable no-useless-computed-key */
```
tbergq, line 3: We could consider setting this rule to warning maybe since flow doesn't work with non-string object properties

src/example-relay/src/components/Navbar.js
```diff
@@ -105,6 +107,18 @@ function Navbar() {
   );
 }
 
+const brandAnimation = sx.keyframes({
+  [0]: {
+    transform: 'translateX(-300px) rotate(0deg)',
+  },
+  [80]: {
+    transform: 'translateX(10px) rotate(-5deg)',
```
tbergq, line 115: Maybe the rotate is too much here 🤷‍♂️

src/example-relay/src/components/Navbar.js
mrtnzlml: Yeah, go for it. Maybe completely turning it off? Could you please offload it in a different PR?

Also, would it be possible to improve the description of the relevant commits in this PR? I'd like to keep the quality level high and this is one thing which I believe should exist. Especially while adding a new feature like keyframes.

src/example-relay/src/components/Navbar.js
mrtnzlml: I'd argue that the whole `brandAnimation` is a bit too much. 😛 How about creating test fixtures inside SX instead? They should be there anyway in my opinion.

src/example-relay/src/components/Navbar.js
mrtnzlml: Not even talking about this typo:

```
xs: Add keyframes support
^^
```

😬 😅

src/example-relay/src/components/Navbar.js
tbergq: Sure, I can remove it entirely 👍 

It was just nice to _prove_ that it works ;)

src/example-relay/src/components/Navbar.js
tbergq: Sorry 🙈 

I will fix it 😂

src/example-relay/src/components/Navbar.js
```diff
@@ -171,6 +185,10 @@ const styles = sx.create({
       outline: 'none',
     },
   },
+  brand: {
+    animation: `${brandAnimation} 1s`,
```
mrtnzlml, line 189: Such string concatenation will make the static analysis quite a bit more challenging when building a Babel plugin for SX. I guess (in this case) it's needed especially because of the shorthand property. Have you considered disallowing this shorthand property OR allowing keyframes only in `animation-name`? Note that shorthand properties are explicitly discouraged in our readme.

tbergq: > Note that shorthand properties are explicitly discouraged in our readme

Yeah, I know, I just think people will use them anyway because it is convenient 🙈

src/sx/README.md
```diff
@@ -195,6 +196,40 @@ const styles = sx.create({
...
+});
+
+const styles = sx.create({
+  text: {
+    animation: `${fadeIn} 2s`,
+  },
+});
+```
+
+It also supports `from and to` for simpler animations.
```
mrtnzlml, line 227:
```suggestion
It also supports `from` and `to` for simpler animations.
```

src/sx/src/create.js
```diff
@@ -16,12 +16,11 @@ type MediaQueries = {|
 |};
 
 // https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes
-type KeyFrames = {|
-  +from: AllCSSPropertyTypes,
-  +to: AllCSSPropertyTypes,
+export type KeyFrames = {|
```
mrtnzlml, line 19: This type was original intended to be used differently (and arguably wrongly). Would it make sense to move it into `src/sx/src/keyframes.js`? That would allow us to clean the types here a little bit.

src/example-relay/src/components/Navbar.js
tbergq: ✔️

src/sx/src/create.js
tbergq: Yeah, sure, I will do so.

tbergq: ✔️

src/sx/README.md
```diff
@@ -227,6 +228,40 @@ const styles = sx.create({
 });
 ```
 
+### Keyframes
+
+SX also has support for keyframes, it exports a function that handles that generates a animation name from the input you give it. You can use it like this:
```
mrtnzlml, line 233:
```suggestion
SX also has support for keyframes, it exports a function that generates an animation name from the input you give it. You can use it like this:
```

src/sx/README.md
```diff
@@ -227,6 +228,40 @@ const styles = sx.create({
...
+    opacity: 0.3,
+  },
+  100: {
+    opacity: 1,
+  },
+});
+
+const styles = sx.create({
+  text: {
+    animation: `${fadeIn} 2s`,
```
mrtnzlml, line 254: Can you please avoid shorthand props? I know it will work but we do not recommend them so it would be nice to avoid it here. What do you think?

src/sx/src/keyframes.js
```diff
@@ -0,0 +1,58 @@
...
+import type { AllCSSPropertyTypes } from './css-properties/__generated__/AllCSSPropertyTypes';
+import styleCollector from './StyleCollector';
+import transformStyleName from './transformStyleName';
+import transformStyleValue from './transformValue';
+import { injectCSS } from './injectRuntimeStyles';
+
+type KeyFrames = {|
+  +from?: AllCSSPropertyTypes,
+  +to?: AllCSSPropertyTypes,
+  +[number]: AllCSSPropertyTypes, // percentages
```
mrtnzlml, line 16: I was thinking that maybe we should change it to string:

```suggestion
  +[string]: AllCSSPropertyTypes, // percentages
```

It's because keyframes can be more complex:

```css
@keyframes identifier {
  0% { top: 0; left: 0; }
  30% { top: 50px; }
  68%, 72% { left: 50px; }
  100% { top: 100px; left: 100%; }
}
```

See: https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes#When_properties_are_left_out_of_some_keyframes

Do you agree?

src/sx/src/keyframes.js
```diff
@@ -0,0 +1,58 @@
...
+|};
+
+const extractStyles = (styles: AllCSSPropertyTypes) => {
+  let cssDefinition = '';
+  for (const [styleName, value] of Object.entries(styles)) {
+    const styleValue: string = (value: any);
```
mrtnzlml, line 22: Can you please avoid the unsound `Object.entries`? This type casting doesn't seem to be correct. The values can be numbers as well for example.

src/sx/src/keyframes.js
tbergq: Yeah, that makes sense 👍

src/sx/src/injectRuntimeStyles.js
```diff
@@ -25,36 +24,62 @@ export default function injectRuntimeStyles(styleBuffer: StyleBufferType) {
...
+    if (cssRule.type === CSSRule.STYLE_RULE) {
+      const styleRule = ((cssRule: any): CSSStyleRule);
+      if (styleRule.selectorText === `.${node.getHash()}`) {
+        return true;
+      }
+    }
+    return false;
+  };
 
   const insertIndex = styleSheet.cssRules.length;
```
tbergq, line 79: If the loop below inserts more than one rule, won't this insert them all at the same index? 🤔

mrtnzlml: That's probably true. Do you have a capacity to fix it?

tbergq: Yeah, I can fix it, different MR though.

mrtnzlml (approved):
👍 good job

PR #1298, opened 2020-11-10
sx: Add note about pseudo classes and specificity

tbergq:


src/sx/README.md
```diff
@@ -142,6 +142,38 @@ const styles = sx.create({
...
+      color: 'pink',
+    },
+    ':active': {
+      color: 'blue',
+    },
+  },
+});
+```
+
+These 2 rules will have the same specificity, and the one defined last in the stylesheet will win. It may or may not help to change order in this style, because the class could be created by a different rule. What will help is to raise the specificity of the active class. You can do this:
```
mrtnzlml, line 160: Do you have a real-world example? Maybe some failing fixture? I am not sure I 100% understand the problem from the description.

tbergq: Check [this codepen](https://codesandbox.io/s/crimson-worker-58t6m?file=/src/styles.css)

Clicking the button doesn't change the color. Then uncomment the commented CSS, and the the color changes when you click the button.

tbergq: Or [this](https://codesandbox.io/s/clever-ardinghelli-n0ry8?file=/src/index.js), since it uses `sx`

mrtnzlml (approved):

PR #1302, opened 2020-11-10
Extract Flow types for Eslint plugins/rules

mrtnzlml:
I need to reuse Flow types for the new `eslint-plugin-sx` so I made them reusable via `@adeira/flow-types-eslint` package. I changed the types a little bit to make them more accurate which resulted in new errors. The idea is to iterate on these types as needed (or replace them if we find something better).


src/flow-types-eslint/src/types/Identifier.js
```diff
@@ -0,0 +1,8 @@
+// @flow strict
+
+import type { INode } from './INode';
+
+export type Identifier = {|
```
tbergq, line 5: `Identifier = $ReadOnly<{|...|}>` since the spread makes it loose the read-only-ness

src/flow-types-eslint/src/types/CallExpression.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import type { INode } from './INode';
+
+// aaa()
+// ^^^^^
+export type CallExpression = {|
```
tbergq, line 7: $ReadOnly because of the spread

src/flow-types-eslint/src/types/ImportDeclaration.js
```diff
@@ -0,0 +1,14 @@
+// @flow strict
+
+import type { INode } from './INode';
+import type { Literal } from './Literal';
+import type { ImportNamespaceSpecifier } from './ImportNamespaceSpecifier';
+import type { ImportSpecifier } from './ImportSpecifier';
+import type { ImportDefaultSpecifier } from './ImportDefaultSpecifier';
+
+export type ImportDeclaration = {|
+  ...INode,
```
tbergq, line 10: Here too

tbergq: There were more of these, so I won't comment on all of them 🙂

src/flow-types-eslint/src/types/Identifier.js
mrtnzlml: Isn't there an Eslint rule for this kind of mistake?

mrtnzlml:
@tbergq Please, have a look again.

src/flow-types-eslint/src/types/Identifier.js
tbergq: Not that I know of, I am not sure there should be one either, because this might be intentional

tbergq (approved):

PR #1304, opened 2020-11-10
Babel SXT: use snapshot testing

mrtnzlml:
Just a proposal. It is a bit cleaner in terms of configuration but might be less readable when you want to study the output code.


michalsanger:
Another small disadvantage could be the way how to provide plugin options. Previously it was just `options.json` in the folder, not has to be in `index.test.js` code.

michalsanger (approved):

mrtnzlml:
@michalsanger That is a good point as well. Feel free to revert it if you find it too annoying to work with. I am building a different Babel plugin and I found it quite convenient. Let's see. :)

PR #1305, opened 2020-11-10
Remove unused 'eslint-plugin-graphql-fragments'

mrtnzlml:
It's one year old without any activity. I asked about it on Slack and got no response. It's also not published to NPM and not used in Universe - essentially a dead project. I'd like to remove it if everyone is OK with that. We can always resurrect it later from Git history.


tbergq (approved):
I am ok with deleting it.

michalsanger (approved):

PR #1316, opened 2020-11-12
sx: Calculate insertIndex inside loop in injectRuntimeStyles

tbergq:
This was defined outside the loop before, it was ok but confusing.
It is easy to think that this works like an array and that you are
overwriting index, though it is just adding and pushing the other down.
Also added a test for that the runtime styles are added correctly

### DIFF
--- a/src/sx/src/__tests__/injectRuntimeStyles.test.js
+++ b/src/sx/src/__tests__/injectRuntimeStyles.test.js
@@ -1,0 +1,67 @@
+// @flow
+
+/* global document */
+
+import { render } from '@testing-library/react';
+
+import * as sx from '../../index';
+
+it('injects runtime styles', () => {
+  render(<style data-adeira-sx />);
+
+  const styleTag = document.querySelector('style');
+  expect(styleTag?.sheet?.cssRules).toHaveLength(0);
+
+  sx.create({
+    test: {
+      opacity: 0,
+      height: '100px',
+    },
+  });
+
+  expect(styleTag?.sheet?.cssRules).toHaveLength(2);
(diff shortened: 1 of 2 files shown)


src/sx/src/__tests__/injectRuntimeStyles.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+  sx.create({
+    test: {
+      opacity: 0,
+      height: '100px',
+    },
+  });
+
+  expect(styleTag?.sheet?.cssRules).toHaveLength(2);
+  expect(styleTag.sheet.cssRules).toMatchInlineSnapshot(`
```
mrtnzlml, line 21: Why are the optional chains required on line 20 and not on line 21? 🤔

mrtnzlml (approved):
Thanks!

src/sx/src/__tests__/injectRuntimeStyles.test.js
tbergq: Yeah, good question, I didn't notice. 

it seems like it was not necessary after adding `globalThis.document`, so I guess that makes the styleTag any?

I will clean it up.

mrtnzlml: Might be. `globalThis` and `global` are type of `any`. Sometimes it's ok to access these global objects directly and just explain it to Eslint like so: `/* global document */`

tbergq: Yeah, that also works 🤔 

My idea was showing lint-error even if I added global document, but I see it passes in the terminal.

PR #1317, opened 2020-11-12
SX Eslint: add new rule 'no-concatenated-classes'

mrtnzlml:
This rule aims to find incorrect usage of template strings inside `className`.

Examples of **incorrect** code for this rule:

```jsx
import { create } from '@adeira/sx';

export default function MyComponent() {
  // Should be:
  //
  //   className={styles('aaa', 'bbb')}
  //
  //   - or -
  //
  //   className={styles('bbb', 'aaa')}
  return <div className={`${styles('aaa')} ${styles('bbb')}`} />;
}

const styles = create({
  aaa: { color: 'red' },
  bbb: { color: 'blue' },
});
```

Examples of **correct** code for this rule:

```jsx
import { create } from '@adeira/sx';

export default function MyComponent() {
  return <div className={styles('aaa', 'bbb')} />;
}

const styles = create({
  aaa: { color: 'red' },
  bbb: { color: 'blue' },
});
```


src/eslint-config-adeira/__tests__/__snapshots__/index.test.js.snap
```diff
@@ -941,5 +944,14 @@ Object {
 
 exports[`rules snapshot: diff of stable and strict rules 1`] = `
 Snapshot Diff:
-Compared values have no visual difference.
+- First value
++ Second value
```
mrtnzlml, line 948: Improved here: https://github.com/adeira/universe/pull/1318

tbergq (approved):

PR #1323, opened 2020-11-13
test-sx/no-unused-stylesheet 

tbergq:


src/example-relay/src/components/Navbar.js
```diff
@@ -106,6 +106,9 @@ function Navbar() {
 }
 
 const styles = sx.create({
+  notUsing: {
```
tbergq, line 109: @mrtnzlml shouldn't this produce a lint error?

mrtnzlml: It should. I am going to check what's wrong.

mrtnzlml: Should be fixed here: https://github.com/adeira/universe/pull/1324

PR #1327, opened 2020-11-14
Eslint runner: add short summary at the end

mrtnzlml:
Looks like this (only when the run succeeded):

```
Skipped files: 65 (no changes)
Checked files: 729
```

This summary is shown only when the lint run was successful because otherwise we are printing the errors/todos.


tbergq (approved):
Nice, looks good 👍

PR #1328, opened 2020-11-14
eslint-runner: Add space after total in progressbar

tbergq:
The last number of the total was cut behind the cursor.
Now you can see the entire total.

Before

Screenshot 2020-11-14 at 09 44 00

After
Screenshot 2020-11-14 at 09 42 00

### DIFF
--- a/src/eslint-runner/reporter.js
+++ b/src/eslint-runner/reporter.js
@@ -43,7 +43,8 @@ class JestProgressBarReporter {
 
   onTestStart() /*: void */ {
     if (this._bar == null) {
-      this._bar = new ProgressBar(':bar :current/:total', {
+      // Note that the space behind :total is necessary so that the cursor doesn't cover the last number of the total.
+      this._bar = new ProgressBar(':bar :current/:total ', {
         complete: '█',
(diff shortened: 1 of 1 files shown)


src/eslint-runner/reporter.js
```diff
@@ -43,7 +43,7 @@ class JestProgressBarReporter {
 
   onTestStart() /*: void */ {
     if (this._bar == null) {
-      this._bar = new ProgressBar(':bar :current/:total', {
+      this._bar = new ProgressBar(':bar :current/:total ', {
```
mrtnzlml, line 46: Aha! I thought it's only my local problem. Could you please add there a simple comment explaining the motivation? Just to make sure we won't "fix it" back.

tbergq: ✅

mrtnzlml (approved):

mrtnzlml:
@tbergq Do you want to release a patch version?

tbergq:
> @tbergq Do you want to release a patch version?

No, I don't need it anywhere else at the moment 😊

PR #1329, opened 2020-11-14
adeira.dev: Add babel.config.js to adeira.dev project

tbergq:
The build was failing after we added the new JSX transform.
This should fix the failing build

### DIFF
--- a/src/monorepo-shipit/config/adeira-dev.js
+++ b/src/monorepo-shipit/config/adeira-dev.js
@@ -9,6 +9,12 @@ module.exports = ({
     };
   },
   getPathMappings() {
-    return new Map([['src/adeira.dev/', '']]);
+    return new Map([
+      ['src/adeira.dev/__github__/babel.config.js', 'babel.config.js'],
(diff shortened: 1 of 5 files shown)


src/monorepo-shipit/config/adeira-dev.js
```diff
@@ -9,6 +9,12 @@ module.exports = ({
...
   },
   getPathMappings() {
-    return new Map([['src/adeira.dev/', '']]);
+    return new Map([
+      ['src/adeira.dev/', ''],
+      ['src/adeira.dev/__github__/babel.config.js', 'babel.config.js'],
+    ]);
+  },
+  getStrippedFiles() {
+    return new Set([/__github__/, /^\.babelrc\.js$/]);
```
mrtnzlml, line 18: Can you please add some tests for it?

mrtnzlml (approved):
The `@adeira/babel-preset-adeira` is not defined in the workspace dependencies. That might be an issue, right? Please have a look at it.

Anyway, thanks for taking care of it! 👍

PR #1330, opened 2020-11-14
SXT: Use Tailwind setup

michalsanger:
This is a fundamental rework of SX Tailwind. Previously the styles definitions for SX were generated from a Tailwind CSS file, now it's generated from the same data the Tailwind itself  is generated.

It opens the way for a full support of customisation and simpler upgrade when a new version of Tailwind is released.


mrtnzlml (approved):
Just a quick read. 🏃

src/sx-tailwind/src/__tests__/sxGenerator.test.js
```diff
@@ -18,41 +22,42 @@ it('works with class selector', () => {
...
           "maxWidth": "640px",
         },
       },
     }
   `);
 });
 
-it('supports nested media queries', () => {
+// eslint-disable-next-line jest/no-disabled-tests
+it.skip('supports nested media queries', async () => {
```
tbergq, line 60: Why are you skipping this test?

src/sx-tailwind/src/__tests__/sxGenerator.test.js
```diff
@@ -163,44 +168,45 @@ it('does not remove rem units from font-size', () => {
...
-it('skips empty styles', () => {
+it('skips empty styles', async () => {
   const css = `.hover\\:empty-style:hover {}`;
 
-  expect(generate(css)).toMatchInlineSnapshot(`Object {}`);
+  expect(await convertToSx(css, config)).toMatchInlineSnapshot(`Object {}`);
 });
 
-it('skips empty media queries', () => {
+it('skips empty media queries', async () => {
```
tbergq, line 194: This is testing keyframes, but the it text says media queries

michalsanger: Renamed to "skips unsupported at-rules"

src/sx-tailwind/src/__tests__/sxGenerator.test.js
michalsanger: It's not implemented yet but I already had this test from previous version.

tbergq (approved):

PR #1332, opened 2020-11-15
Eslint SX: improve error nodes location

mrtnzlml:
Previously, we were reporting the errors with 'Program' node which resulted in the whole file being marked as errored. That's a bad practice since you need to go and search for the actual error location. This commit improves the behavior by reporting the errors with relevant nodes. This results in much better error location reporting.


tbergq:
> Previously, we were reporting the errors with 'Program' node which resulted in the whole file being marked as errored.

That's why the error was weird 🙈

tbergq (approved):

PR #1334, opened 2020-11-15
Remove node 12.x from node-version matrix

tbergq:
Node 12 is no longer lts version, it is now Node 14 that is lts

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [12.x, 14.x, 15.x]
+        node-version: [14.x, 15.x]
 
     steps:
       - uses: actions/checkout@v1
--- a/.github/workflows/npm-publisher.yml
+++ b/.github/workflows/npm-publisher.yml
@@ -15,10 +15,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v1
-      - name: Use Node.js 12.x
+      - name: Use Node.js 14.x
         uses: actions/setup-node@v1
         with:
-          node-version: 12.x
+          node-version: 14.x
       - name: publish NPM packages to the registry
         run: |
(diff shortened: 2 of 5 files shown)


michalsanger (approved):

mrtnzlml:
@tbergq Isn't it a bit early? Version 12 is still active LTS (https://nodejs.org/en/about/releases/). I had it scheduled for NOV 30 the earliest.

Moreover, now we have to cut major versions or at least update the chnagelogs everywhere. Have you thought about that? 🤔

tbergq:
> @tbergq Isn't it a bit early? Version 12 is still active LTS (https://nodejs.org/en/about/releases/). I had it scheduled for NOV 30 the earliest.
> 
> Moreover, now we have to cut major versions or at least update the changelogs everywhere. Have you thought about that? 🤔

#1336

mrtnzlml:
@tbergq How are we going to proceed when it's not active LTS anymore? I was thinking to update the changelogs and leave there a message for us to decide later if the major bump is needed for the particular package or not. What do you think?

tbergq:
> @tbergq How are we going to proceed when it's not active LTS anymore? I was thinking to update the changelogs and leave there a message for us to decide later if the major bump is needed for the particular package or not. What do you think?

I actually think we can keep running the CI in 12.x, for a while longer. I see other OSS still running their CI in Node 10. Then when we decide to remove the test run in 12.x we can drop a message about it in the changelogs I guess. 

For now, we could just change our deploy jobs and such to run in 14.x.

mrtnzlml:
> Then when we decide to remove the test run in 12.x we can drop a message about it in the changelogs I guess.

Agree. Something like "Node.js version 12 is no longer supported - it might continue working but we highly recommend upgrading" could be enough. I have to think about the major bumps though - at least for some packages.

> For now, we could just change our deploy jobs and such to run in 14.x.

True, I am going to do it. Especially OSS versions are running outdated versions.

PR #1341, opened 2020-11-16
Eslint SX: automatically include only JS files

mrtnzlml:
Previously, we included also '.js.flow' files which are present in the NPM exported version. This commit should prevent that by including only JS files. I also disabled generation of MJS files which are not needed for this Eslint plugin.


src/eslint-plugin-sx/src/getRules.js
```diff
@@ -11,7 +11,7 @@ module.exports = function getRules() /*: {| +[ruleName: string]: string |} */ {
     withFileTypes: true,
   })) {
     // $FlowIssue[prop-missing]: https://github.com/facebook/flow/pull/8532
-    if (dirent.isFile()) {
+    if (dirent.isFile() && /\.js$/.test(dirent.name)) {
```
tbergq, line 14: Would it be possible to test this somehow, in case someone tries to change it and possibly breaks it again?

tbergq (approved):

src/eslint-plugin-sx/src/getRules.js
mrtnzlml: Added tests, please have a look again. :)

tbergq (approved):

PR #1352, opened 2020-11-16
SX: change imports/exports to allow composability

mrtnzlml:
I would like to propose a change of the default export which would allow us to merge external SX stylesheets. For example, here is a base component:

```
import sx from '@adeira/sx';

type Props = {|
  +xstyle: {|
    +marginTop: number,
  |},
|};

const styles = sx.create({ default: { fontSize: 16 } });

function MyBaseComponent(props: Props) {
  return <div className={sx(styles.default, props.xstyle)} />;
}
```

And another wrapping component which is sending down some custom styles:

```
const styles = sx.create({ spacing: { marginTop: 4 } });

function MyCustomComponent() {
  return <MyBaseComponent xstyle={styles.spacing} />;
}
```

We are currently unable to achieve such behavior without this change.

This change is inspired by StyleX, see: https://youtu.be/9JZHodNR184

### DIFF
--- a/src/babel-plugin-transform-sx-tailwind/README.md
+++ b/src/babel-plugin-transform-sx-tailwind/README.md
@@ -7,7 +7,7 @@ This has a positive impact on the final bundle size because the huge Tailwind de
 ```diff
 import React from 'react';
 - import { tailwind } from '@adeira/sx-tailwind';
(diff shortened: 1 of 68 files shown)


tbergq (approved):
LGTM

PR #1354, opened 2020-11-17
SXT Website: Showcase with source code snippets

michalsanger:
Previous syntax highlighter replaced with Refractor because of better API and 5x smaller bundle size.


mrtnzlml (approved):
It would be super nice if you could find a way how to extract the actually used code from the components.

michalsanger:
The real code can have too much noise, I'd rather define the highlighted source code manually.

PR #1357, opened 2020-11-17
Dependencies: upgrade to Relay version 10.1.0

mrtnzlml:


tbergq (approved):
Should we release a new version of `@adeira/relay`-packages as well?

mrtnzlml:
@tbergq Can be. I bumped the affected packages. 🙂

tbergq (approved):

PR #1358, opened 2020-11-17
SX eslint: validate 'keyframes' usage

mrtnzlml:
The validation is basically the same with `sx.create` so I made the logic more generic and added tests for keyframes as well.


src/eslint-plugin-sx/docs/rules/no-unused-stylesheet.md
```diff
@@ -53,4 +53,4 @@ _none_
 
 ## When Not To Use It
 
-TKTK
+There should be no valid reason to turn this rule off. It helps with a dead code elimination.
```
tbergq, line 56: > When Not To Use It

`If you are not using sx` instead maybe?

tbergq (approved):

src/eslint-plugin-sx/docs/rules/no-unused-stylesheet.md
mrtnzlml: Well, it's inside eslint-plugin-sx so that would relate to the whole package. Similarly "if you are not using Eslint". :) In my eyes this rather describes the single rule usage when already using SX with this plugin.

tbergq: Good point 😂

PR #1359, opened 2020-11-17
Bazel CI: offload some minor changes

mrtnzlml:
- disable Buildifier lint (too slow without cache)
- offload missing .npmignore changes
- offload missing SX test dependencies
- add ignores into flowconfig


.flowconfig
```diff
@@ -1,7 +1,10 @@
 ; https://flow.org/en/docs/config/ignore/
 [ignore]
 <PROJECT_ROOT>/\.git/.*
-<PROJECT_ROOT>/src/docs/build/
+<PROJECT_ROOT>/bazel-bin/.+
+<PROJECT_ROOT>/bazel-out/.+
+<PROJECT_ROOT>/bazel-testlogs/.+
+<PROJECT_ROOT>/bazel-universe/.+
```
tbergq, line 7: Does this work for you? Because running flow for me takes forever. I think flow has issues with resolving these, since they are symlinks. I have to add `/private/var/tmp/.*` it ignore for it to work as normal.

mrtnzlml: @tbergq Please, have a look again. I offloaded it as it is here: https://github.com/adeira/universe/pull/1068/files (previously I changed it for this PR without even trying it 🙈)

tbergq: Seems to work now 👍

tbergq (approved):

PR #1364, opened 2020-11-18
SXT Babel: Utility name suggestion

### LINKED ISSUE #1288: SX: Improve suggested stylesheet name
When an invalid classname is used like here:
```jsx
<button
  className={tailwind('bg-blue-500 hover:bg-blue-700 text-white font-bold py-2 px-4 rounde')}
  type="button"
>
  Button
</button>
```

The error is thrown:
`Invariant Violation: SX was called with 'rounde' stylesheet name but it doesn't exist. Did you mean 'py-2' instead?`

Which is not really helpful, we could do better:
`Invariant Violation: SX was called with 'rounde' stylesheet name but it doesn't exist. Did you mean 'rounded' instead?`

SX takes suggestions only from the current `sx.create` call, but for SX Tailwind where we have a limited set of classnames, this set could be used for suggestions.

Could `@adeira/sx` provide something like `registerSuggestions(classNames: Set<string>): void`?

michalsanger:
When there is an error because of invalid utility name, the message contains a suggestion to the correct name.

Closes #1288


yarn.lock
```diff
@@ -9,6 +9,22 @@
   dependencies:
     "@babel/runtime" "^7.11.2"
 
+"@adeira/sx-tailwind@^0.8.0":
```
mrtnzlml, line 12: This suggests that SX Tailwind was not upgraded everywhere in the Universe to the latest version. Could you check it?

mrtnzlml (commented):
The original issue can be closed after this PR, right?

yarn.lock
michalsanger: Fixed

mrtnzlml (approved):

PR #1370, opened 2020-11-18
SXT Website: Snippets for gradients and transitions

michalsanger:


mrtnzlml (approved):
I really like how the Showcase look like 👍

PR #1380, opened 2020-11-20
Eslint SX: add support for error locations

mrtnzlml:
Originally I wanted to fix some other issue but that turned out to be too complicated. So I am offloading at least the improved error locations (and some additional test cases).


src/eslint-plugin-sx/src/rules/__tests__/fixtures/no-unused-stylesheet/invalid/unrelated-call-expression.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @flow
+ * @eslintExpectedError SX function "styles" was not used anywhere in the code.
```
michalsanger, line 3: Why is the location missing here?

mrtnzlml: It's an opt-in feature and I didn't want to add it everywhere. 🙂

michalsanger (approved):

Issue #932, opened 2020-08-17
Migrate from dependabot-preview to native Github Dependabot


mrtnzlml:
Essentially set up this: https://github.com/adeira/universe/network/updates

Note: native Github Dependabot doesn't support automerge option. This would be handy even for other PRs in the queue so we should set up generic automerging first. Some options:

- https://github.com/apps/probot-auto-merge
- https://github.com/palantir/bulldozer
- https://alexwlchan.net/2019/03/creating-a-github-action-to-auto-merge-pull-requests/

PR #1400, opened 2020-11-24
CI: explicitly setup Dependabot

mrtnzlml:
Currently, we are using `dependabot-preview` which supports automatic merges but it's not configurable anymore. This commit migrates it to native GitHub Dependabot which allows us to configure other automated updates like GitHub actions. Automatic merging is done via Kodiak.

Closes: https://github.com/adeira/universe/issues/932

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+
+updates:
+  - package-ecosystem: 'npm' # See documentation for possible values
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
...
+version: 2
+
+updates:
+  - package-ecosystem: 'npm' # See documentation for possible values
+    directory: '/' # Location of package manifests
+    schedule:
+      interval: 'daily'
+    labels:
+      - 'dependencies'
+      - 'automerge' # see Kodiak `merge.automerge_label`
```
michalsanger, line 15: Does it mean Kodiak will automatically merge the PR when all required checks are OK?

mrtnzlml: Yes, which includes at least one approval from us, see: https://kodiakhq.com/docs/recipes#configuring-dependabot-with-the-automerge-label

michalsanger (approved):

PR #1405, opened 2020-11-25
SX Tailwind: Use Tailwind CSS v2.0

michalsanger:
The new major version of Tailwind CSS now depends more on configuration. For example the default colors palette was reduced and shifted a little bit, users should define colors explicitly in the config. CSS variables are prefixed with `tw` now and some utilities were renamed.

The main change in SX Tailwind is in the Babel plugin. Styles are now generated and cached based on the configuration so every utility should be configurable. Previously only font family was.


mrtnzlml (approved):
The changes in `src/babel-plugin-transform-sx-tailwind/src/index.js` feel a bit hacky. 😬

michalsanger:
I feel the same, unfortunately the PostCSS that generates Tailwind classes is async and Babel is synchronous. 

Maybe some preBuild hook could help? I guess handle all edge cases in tests and watch mode will not be a simple job. I'll see...

Issue #1301, opened 2020-11-10
adeira/universe whishlist


mrtnzlml:
_This [megathread](https://www.reddit.com/r/OutOfTheLoop/comments/2yuaad/what_is_a_megathread/?utm_source=share&utm_medium=web2x&context=3) is about `adeira/universe` itself._

## Markdown TOC generator

GitHub markdown implementation doesn't support an automatic table of content generation. However, I find it useful sometimes so it would be nice if we could have some alternative (maybe with the ability to automatically send PRs).

## Bazel build system

We want to support other languages, not just JavaScript. Moreover, we are sometimes hitting limitations of Yarn Workspaces (see our `@babel/register` hack). Bazel should resolve all these issues.

## Automatically generate `.eslintignore` and `.flowconfig` files from the child configs

These files can be only in the root but that's a bit strange in a monorepo context. It would be nice if every project could define these configs and we automatically merge them together into one superconfig. Again, would be nice to have it completely automated including the PRs submission (so we don't have to think about it).

---

_Already resolved tasks:_

## ✅ Scanner for `@adeira` dependencies in `yarn.lock` (https://github.com/adeira/universe/pull/1421)

We should not have any `@adeira` dependencies inside `yarn.lock`. Such a situation suggests that we didn't upgrade some dependency internally. It would be nice to catch such a mistake on CI. Possible solution: https://github.com/yarnpkg/yarn/tree/a4708b29ac74df97bac45365cba4f1d62537ceb7/packages/lockfile

## ✅ Automatic merge train (https://github.com/adeira/universe/pull/1391)

It would be very handy to have something like "automatic merge train" where you could just approve a PR and it would be automatically rebased and merged when possible (even if there are many other PRs in the queue).

PR #1419, opened 2020-11-27
Deprecate '@adeira/test-utils'

mrtnzlml:
I was thinking for a while what to do with `@adeira/test-utils` package. There are only two functions (`evaluateGraphQLResolver` and `generateTestsFromFixtures`) the GraphQL resolver evaluation seems to be pretty useless/outdated. On the other hand the fixtures test generator is quite popular and helpful.

I'd like to propose a deprecation of https://www.npmjs.com/package/@adeira/test-utils and I'd like to create a new NPM package instead which is specializing only on the fixtures testing.


src/graphql-resolve-wrapper/package.json
```diff
@@ -4,15 +4,12 @@
...
   "private": false,
-  "version": "0.2.0",
+  "version": "0.3.0",
   "main": "src/index",
   "sideEffects": false,
   "dependencies": {
     "@babel/runtime": "^7.12.5"
   },
-  "devDependencies": {
-    "@adeira/test-utils": "^0.2.0"
```
mrtnzlml, line 14: Even we were using some prehistoric version of the GraphQL resolver tester.

yarn.lock
```diff
@@ -2,20 +2,6 @@
...
-  dependencies:
-    "@babel/runtime" "^7.11.2"
-
-"@adeira/test-utils@^0.2.0":
-  version "0.2.0"
-  resolved "https://registry.yarnpkg.com/@adeira/test-utils/-/test-utils-0.2.0.tgz#9e9feef7bb10968d161dcd4a568e760af25b7c50"
-  integrity sha512-RhnDtUQMNmhzQyGtjewXckGtBvn15qNOnNrIy5fnLTAY1vjowjglKpB+AXl2eq7xpkf+2k6R4J7v06NYU2HmQw==
-  dependencies:
-    "@adeira/js" "^1.2.0"
-
```
mrtnzlml, line 18: No `@adeira` packages should ever be in `yarn.lock`.

tbergq: Could we add some scanner test for this maybe?

tbergq (approved):

yarn.lock
mrtnzlml: Added into wishlist: https://github.com/adeira/universe/issues/1301

PR #1421, opened 2020-11-28
scanner: Test that we have no @adeira package in yarn.lock

tbergq:
If we forget to upgrade a dependency in one of our packages
We might get an old version from npm installed instead.
This should fail the CI if we don't keep our deps up to date


mrtnzlml (approved):

mrtnzlml:
@tbergq I just realized: what if it's a dependency of some other package? 🤔

tbergq:
> @tbergq I just realized: what if it's a dependency of some other package? 🤔

Yeah, that would probably make it fail, but I think we already had a problem when that happened when we use `@kiwicom/orbit-components` and that used `@adeira/js`, that made our `relay-example` build fail if I remember correctly.

PR #1425, opened 2020-11-30
Setup .flowconfig files for React Native

mrtnzlml:
This commits adds support multiple .flowconfig versions for React Native (iOS, Android). It's fully backward compatible and should have minimal impact on CI performance (seems like the started server is being reused + there is a saved state from the previous run).


mrtnzlml:
TODO: check the CI - it behaves fishy 🤔

tbergq (approved):

mrtnzlml:
Just a note: it's not true that the saved state is being reused between iOS and Android. Flow server restarts with the following message:

```text
Invalid saved state: .flowconfig has changed since this saved state was generated.
Failed to load saved state: .flowconfig has changed since saved state was generated
```

PR #1427, opened 2020-11-30
Bootstrap 'src/ya-comiste-rust' backend

mrtnzlml:
I am still working on the schema design so this is just a project bootstrap so I can setup all the things around it.


mrtnzlml:
Currently failing on this issue: https://github.com/facebook/jest/issues/8922 (in our case not a Linux `*.snap` but snapshot in Rust code).

tbergq (approved):
I guess I have to learn rust also no 😂 

Not really into that language yet, but looks good as far as I can tell 🤷‍♂️

PR #1461, opened 2020-12-05
monorepo-test-runner: run all tests if a config file is changed

### LINKED ISSUE #1433: monorepo-run-tests not detecting file change in the root workspace
Hi, is there a way to set all tests to run when a file in root workspace changes. 

e.g when you change something in `setupTest.js` which is set in `jest.config.js`
Making a change in setupTest.js does not trigger all tests, which it should in my opinion. 
```
const commonProjectConfig = {
  globals: {
    __DEV__: true,
  },
  timers: "fake",
  setupFilesAfterEnv: [path.join(__dirname, "scripts", "setupTest.js")],
  transform: {
    "^.+\\.[jt]sx?$": "babel-jest",
  },
};
```

tbergq:
Changes in .jest.config.js will trigger all tests to run.
The test runner will also load the config file and check if any
files in setupFilesAfterEnv or setupFiles has changed. If so
all tests will be executed.

 Release @adeira/monorepo-utils 0.10.0
    
 closes #1433


mrtnzlml:
> The initial idea was to require .jest.config.js and look for setup files,
but that fails because babel tries to transpile DEV
which generates syntax error.

@tbergq Can you elaborate on this? What was the issue exactly? Is it because of the `global.__DEV__ = true` hack we use there? 🤔

tbergq:
> Is it because of the `global.__DEV__ = true` hack we use there?

Yes, from what I understand from the error message, it is because it tries to transpile this into `global.process.env.NODE_ENV !== 'producition' = true`

mrtnzlml:
Yeah, that makes sense but it's a pity because the original solution sounds much better. I'd like to propose a different solution. What about this?

```patch
Index: package.json
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/package.json b/package.json
--- a/package.json	(revision e565bb05667710c548d5cac37da82bc68a6813e6)
+++ b/package.json	(date 1607188087236)
@@ -33,7 +33,7 @@
     "lint": "eslint --config=.eslintrc.js --print-config=babel.config.js | yarn run eslint-config-prettier-check && yarn run prettier --check \"src/**/*.md\" && yarn run jest --config=.jest-eslint.config.js",
     "scanner": "yarn run jest --config src/monorepo-scanner/.jest.config.js",
     "test": "yarn run typecheck && yarn run lint && yarn run test-only --ci --colors && yarn run scanner",
-    "test-only": "BABEL_TYPES_8_BREAKING=true yarn run monorepo-babel-node src/monorepo-utils/bin/monorepo-run-tests.js",
+    "test-only": "src/monorepo-utils/bin/monorepo-run-tests.js",
     "typecheck": "yarn run typecheck-ios && yarn run typecheck-android",
     "typecheck-ios": "yarn run adeira-flow-bin --flowconfig-name=.flowconfig.ios",
     "typecheck-android": "yarn run adeira-flow-bin --flowconfig-name=.flowconfig.android"
Index: src/monorepo-utils/bin/monorepo-run-tests.js
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/monorepo-utils/bin/monorepo-run-tests.js b/src/monorepo-utils/bin/monorepo-run-tests.js
--- a/src/monorepo-utils/bin/monorepo-run-tests.js	(revision e565bb05667710c548d5cac37da82bc68a6813e6)
+++ b/src/monorepo-utils/bin/monorepo-run-tests.js	(date 1607188275172)
@@ -2,6 +2,18 @@
 
 // @flow
 
+// See: https://jestjs.io/docs/en/cli#--configpath
+console.warn(
+  JSON.stringify(require('/Users/martinzlamal/Work/adeira/universe/.jest.config.js'), null, 2),
+);
+
+// BEGIN-ADEIRA-UNIVERSE-INTERNAL
+require('@babel/register')({
+  ignore: [/node_modules\/(?!@adeira)/],
+  rootMode: 'upward',
+});
+// END-ADEIRA-UNIVERSE-INTERNAL
+
 const TestsRunner = require('../src/TestsRunner');
 
 const externalConfig = process.argv.slice(2);
```

Essentially, do not transpile the `monorepo-run-tests.js` right away (since it's not needed) but postpone it when it's actually needed. What do you think?

tbergq:
> Essentially, do not transpile the monorepo-run-tests.js right away (since it's not needed) but postpone it when it's actually needed. What do you think?

Yes, that might work. I think the only assumption would be that tests are being run from the root. So I can find the file like this `require(path.join(process.cwd(), '.jest.config.js'))`, since I cannot use `findMonorepoRoot()` before babel-register is loaded.

tbergq:
@mrtnzlml please check again.

mrtnzlml (approved):

mrtnzlml:
Thanks! 👍

Issue #996, opened 2020-08-29
SX: ideas for improvement and further development


mrtnzlml:
This is a collection of ideas we could consider implementing in `@adeira/sx`. It's not certain whether every idea is a good idea though so **do not** consider this to be a roadmap.

Legend:

🤔 - needs further investigation
🧑‍💻 - we should develop this
:construction: - under development

## :construction: Autoprefixer based on Browserlist info (https://github.com/adeira/universe/pull/1467)

Motivation: automatically prefix some problematic properties for older browsers. Inspiration:

- https://github.com/mdn/browser-compat-data/tree/master/css/properties (contains info about prefixes and browser versions)
- https://github.com/robinweser/inline-style-prefixer
- https://www.npmjs.com/package/stylis
- https://github.com/postcss/autoprefixer

## 🧑‍💻 Babel transpilation

Motivation: transpile out code we can to reduce the production bundle size and overall performance. We could print the final atomic CSS classes directly into React markup.

## 🤔 SX function could accept an object with booleans

At this moment, we allow the following syntax for conditional styles:

```js
styles('blue', isRed && 'red')
```

We could also allow this:

```js
styles({
  blue: isBlue,
  red: isRed
});
```

See: https://youtu.be/9JZHodNR184

## 🤔 Additional Eslint rules for `sx()`, `sx.create()` and `sx.keyframes()`

Some ideas:

- warn on shorthand properties (? - yields larger and less optimized output)
- lint of `sx.keyframes()` (we do not have any rules for it atm)

## 🤔 Improve DX when in `__DEV__`

Motivation: it's quite difficult to work with atomic CSS classes when developing. Instead, we could generate developer-friendly output in dev and atomic classes only in production.

## 🤔 Support for `marginEnd`, `marginVertical`, ...

Motivation: easy support for RTL/LTR layouts, see https://reactnative.dev/docs/layout-props#marginend

```text
marginHorizontal -> ['marginRight', 'marginLeft']
marginVertical -> ['marginTop', 'marginBottom']
paddingHorizontal -> ['paddingRight', 'paddingLeft']
paddingVertical -> ['paddingTop', 'paddingBottom']
```

## 🤔 Utility `sx.inline`

Motivation: to be able to define a single stylesheet instead of using `sx.create` and defining only one named stylesheet there.

```jsx
export default function Example() {
  return (
    <div className={sx.inline({ fontSize: 32, padding: 0, textDecoration: 'none' })}>example</div>
  );
}
```

[...]

PR #1467, opened 2020-12-06
SX: Vendor prefixes via Autoprefixer

michalsanger:
Vendor prefixes are automatically added into the generated CSS.

### DIFF
--- a/src/sx/package.json
+++ b/src/sx/package.json
@@ -4,7 +4,7 @@
   "homepage": "https://github.com/adeira/universe/tree/master/src/sx",
   "license": "MIT",
   "private": false,
-  "version": "0.21.0",
+  "version": "0.22.0",
   "main": "index",
   "sideEffects": false,
   "module": false,
@@ -14,11 +14,13 @@
     "@adeira/murmur-hash": "^1.0.0",
     "@adeira/signed-source": "^1.0.0",
     "@babel/runtime": "^7.12.5",
+    "autoprefixer": "^10.0.4",
     "change-case": "^4.1.2",
     "css-tree": "^1.1.2",
     "fast-levenshtein": "^3.0.0",
     "json-stable-stringify": "^1.0.1",
     "mdn-data": "^2.0.14",
+    "postcss": "^8.1.14",
     "prettier": "^2.2.1"
   },
   "peerDependencies": {
--- a/src/sx/src/__tests__/__snapshots__/fixtures.test.js.snap
+++ b/src/sx/src/__tests__/__snapshots__/fixtures.test.js.snap
@@ -36,8 +36,7 @@ exports[`matches expected output: @supports.js 1`] = `
     color: #f00;
   }
 }
-@supports (transform-style: preserve) or (-moz-transform-style: preserve) or
-  (-o-transform-style: preserve) or (-webkit-transform-style: preserve) {
+@supports (transform-style: preserve) or (-webkit-transform-style: preserve) {
   ._2gYAIc._2gYAIc {
     color: #f00;
   }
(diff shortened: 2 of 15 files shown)


tbergq:
Does it work with runtime styles as well?

src/sx/package.json
```diff
@@ -14,11 +14,13 @@
     "@adeira/murmur-hash": "^1.0.0",
     "@adeira/signed-source": "^1.0.0",
     "@babel/runtime": "^7.12.5",
+    "autoprefixer": "^10.0.4",
```
tbergq, line 17: Will this be a part of the runtime-bundle? because it is quite huge

https://bundlephobia.com/result?p=autoprefixer@10.0.4

src/sx/package.json
```diff
@@ -14,11 +14,13 @@
...
     "change-case": "^4.1.2",
     "css-tree": "^1.1.2",
     "fast-levenshtein": "^3.0.0",
     "json-stable-stringify": "^1.0.1",
     "mdn-data": "^2.0.14",
+    "postcss": "^8.1.14",
```
tbergq, line 23: This also adds about 60kb

src/sx/package.json
michalsanger: It's used in https://github.com/vercel/next.js/tree/canary/examples/with-tailwindcss do you see any issue there?

tbergq: This is production build of relay-example: 

Screenshot 2020-12-06 at 14 52 35

tbergq: And this is current master:

Screenshot 2020-12-06 at 14 54 54

src/sx/src/__tests__/__snapshots__/fixtures.test.js.snap
```diff
@@ -36,8 +36,7 @@ exports[`matches expected output: @supports.js 1`] = `
     color: #f00;
   }
 }
-@supports (transform-style: preserve) or (-moz-transform-style: preserve) or
-  (-o-transform-style: preserve) or (-webkit-transform-style: preserve) {
```
mrtnzlml, line 40: So it also removes unneeded prefixes? Cool! 🙂

This makes me think though: shouldn't we specify some `.browserslistrc` explicitly for these tests? Which versions are they using now? Isn't it a bit flaky now?

michalsanger: We could but it would probably have to be in the root folder. 

Or is there a way how we can pass env vars to some tests? https://github.com/browserslist/browserslist#environment-variables

tbergq: > Or is there a way how we can pass env vars to some tests?

You can use https://jestjs.io/docs/en/tutorial-react-native#setupfiles to set env vars for test files.

src/sx/package.json
michalsanger: It is true that during a build of `sx-tailwind-website` in the monorepo I get warnings and increased bundle size:
```
/Users/michalsanger/projects/adeira/universe/node_modules/@adeira/sx/node_modules/browserslist/node.js
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

/Users/michalsanger/projects/adeira/universe/node_modules/@adeira/sx/node_modules/browserslist/node.js 175:18-177:7
Critical dependency: the request of a dependency is an expression
```

But when I tried `postcss` and `autoprefixer` on a standalone Next.js project everything was fine. Maybe because of `next.config.js`?

tbergq: I guess it depends on how it is used. In our case, I am pretty sure using it in the stylecollector will make it a part of the runtime bundle, and I would prefer some duplicated CSS properties in my stylesheet over a +1MB addition to the runtime bundle 😬

src/sx/src/__tests__/__snapshots__/fixtures.test.js.snap
michalsanger: Looks like `process.env.BROWSERSLIST = 'last 2 versions';` in `jest.config.js` did that job

src/sx/package.json
mrtnzlml: So is there some cheaper alternative for CSS-in-JS? 1 MB is really a lot. Why is it so big? 🤔

michalsanger: Maybe the PostCSS is designed to be used with Babel & Webpack, not during a runtime. I'll try to use https://github.com/robinweser/inline-style-prefixer in a new PR but probably the output will not be that good.

tbergq: I think most `CSS-in-JS` uses [styleis](https://www.npmjs.com/package/stylis)

mrtnzlml: Alternatively, you could read it from `mdn/browser-compat-data` as suggested here: https://github.com/adeira/universe/issues/996

michalsanger: Attempt with Stylis #1480

michalsanger:
See #1480

mrtnzlml:
@michalsanger So this branch can be deleted? 🙂

PR #1473, opened 2020-12-07
Shipit: improve logging

mrtnzlml:
This commit hides console outputs from Git and instead displays custom messages for more understandable output about what's going on inside Shipit. Similar to how original FBShipit works: https://github.com/facebook/fbshipit/blob/a3c7cd8c97ef1b49c2b9c371b14b0051c3a7631b/DEBUGGING.md

### DIFF
--- a/src/monorepo-shipit/src/Changeset.js
+++ b/src/monorepo-shipit/src/Changeset.js
@@ -1,4 +1,7 @@
-// @flow strict
+// @flow
+
+import { sprintf } from '@adeira/js';
+import logger from '@adeira/logger';
 
 type Diff = {|
   +path: string,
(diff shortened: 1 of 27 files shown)


src/monorepo-shipit/src/Changeset.js
```diff
@@ -78,6 +83,19 @@ export default class Changeset {
     return this.__clone({ diffs });
   }
 
+  withDebugMessage(string: string, ...args: $ReadOnlyArray<string>): Changeset {
+    const messages = this.debugMessages ?? [];
+    messages.push(sprintf(string, ...args));
```
tbergq, line 88: This will push also into `this.debugMessages` since you are assigning the reference of this.debugMessages to const messages.

Not sure if that is what you want here. Did you mean to do `const messages = [...(this.debugMessages ?? [])]` instead?

tbergq (approved):

src/monorepo-shipit/src/Changeset.js
mrtnzlml: Changed using non-mutating `concat`, thanks! :)

PR #1474, opened 2020-12-07
YCR: implement database connection pool

mrtnzlml:
Opening a new database connection every time one is both inefficient and can lead to resource exhaustion under high traffic conditions. A connection pool maintains a set of open connections to a database, handing them out for repeated use.

See: https://github.com/bikeshedder/deadpool


tbergq (approved):
LGTM, is there also some idle timeout, so open unused connections are not left consuming resources?

mrtnzlml:
> LGTM, is there also some idle timeout, so open unused connections are not left consuming resources?

Nothing I am aware of. The connection pool lives and dies with a request (similarly to how dataloaders would be in a GraphQL context) so that should not be an issue.

PR #1480, opened 2020-12-08
SX: Automatic vendor prefixes

michalsanger:
Vendor prefixes are automatically added into the generated CSS.

### DIFF
--- a/src/sx/src/injectRuntimeStyles.js
+++ b/src/sx/src/injectRuntimeStyles.js
@@ -3,6 +3,7 @@
 /* global document */
 
 import { invariant, warning } from '@adeira/js';
+import { compile, serialize, stringify, prefixer, middleware } from 'stylis';
 
 import StyleCollectorAtNode from './StyleCollectorAtNode';
 import StyleCollectorNode from './StyleCollectorNode';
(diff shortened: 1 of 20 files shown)


src/sx/src/injectRuntimeStyles.js
```diff
@@ -82,16 +83,25 @@ export default function injectRuntimeStyles(styleBuffer: StyleBufferType) {
     if (node instanceof StyleCollectorNode) {
       if (hasStyleRule(matchFunction(node)) === false) {
         // apply missing styles
-        styleSheet.insertRule(node.printNodes().join(''), insertIndex);
+        const rule = serialize(
+          compile(node.printNodes({ trailingSemicolon: true }).join('')),
```
tbergq, line 87: Can't comment on that line, but you probably need this for line 62 as well, inject runtime keyframes.

michalsanger: Done

tbergq (approved):

src/sx/src/injectRuntimeStyles.js
```diff
@@ -58,7 +59,10 @@ export function injectRuntimeKeyframes(css: string, name: string) {
   };
 
   if (hasStyleRule(matchFunction) === false) {
-    styleSheet.insertRule(css, styleSheet.cssRules.length);
+    const rules = compile(serialize(compile(css), middleware([prefixer, stringify]))) ?? [];
+    rules.forEach((rule) => {
+      styleSheet.insertRule(serialize([rule], stringify), styleSheet.cssRules.length);
```
mrtnzlml, line 64: I don't want to discourage you from these efforts but at the same time, I don't want to leave this without mentioning my concern: isn't `serialize(compile(serialize(compile(css))))` a bit too much? 🤔 Have you thought about this @michalsanger?

michalsanger: Yes that was ugly, the api of Stylis is a little bit strange. Please check again.

mrtnzlml (approved):

PR #1505, opened 2020-12-11
Offload Ya Comiste changes

mrtnzlml:
The idea behind this PR is to bulk together multiple Ya Comiste commits knowing that doing code-review for Rust code might be difficult. I am fine to get a review once in a while and merge it in a bulk. 😉


src/ya-comiste-react-native/src/relay/relayEnvironment.js
```diff
@@ -1,15 +1,30 @@
...
 
-function fetchQuery(operation, variables) {
-  // TODO: specify JSON.schema
-  if (variables.entrypointID === 'com.yaComiste.Explore') {
-    return require('./mocks/com.yaComiste.Explore.json');
-  } else if (variables.entrypointID === 'com.yaComiste.ExploreDetail') {
-    return require('./mocks/com.yaComiste.ExploreDetail.json');
+export const handleData = (response: Response): Promise<$FlowFixMe> | Promise<string> => {
+  const contentType = response.headers.get('content-type');
+  if (contentType != null && contentType.indexOf('application/json') !== -1) {
```
tbergq, line 8: doesn't `headers.get` return string or null? Why not just check contentType === 'application/json'?

mrtnzlml: This function was unused (copy-pasted by accident) so I completely removed it. 🙂

tbergq (approved):

PR #1507, opened 2020-12-13
@adeira/logger: remove Datadog support

mrtnzlml:
This logger is quite prehistoric code and we don't use it anymore the way it was designed for. I am considering to remove it completely but before doing so I am at least reducing it to absolute minimum (there are still valid use-cases in Node.js code).


michalsanger (approved):

mrtnzlml:
I just noticed there are many (thousands) weekly downloads on NPM: https://www.npmjs.com/package/@adeira/logger That's quite suspicious. 🤔 Let's see if someone lets us know we broke it in which case I'd reconsider this PR.

tbergq:
> I just noticed there are many (thousands) weekly downloads on NPM: https://www.npmjs.com/package/@adeira/logger That's quite suspicious. 🤔 Let's see if someone lets us know we broke it in which case I'd reconsider this PR.

`@adeira/relay` has `@adeira/logger` as a dependency, that is probably why it is being downloaded.

mrtnzlml:
@tbergq If that's the reason then there is nothing to worry about. :)

PR #1508, opened 2020-12-13
Tests runner: remove unused timezones support

mrtnzlml:
Legacy code from times when we were using GitLab. It's easily replaceable by using `TZ` env variable and calling the runner several times.


src/monorepo-utils/bin/monorepo-run-tests.js
```diff
@@ -12,23 +12,19 @@ require('@babel/register')({
...
 // END-ADEIRA-UNIVERSE-INTERNAL
+
 const TestsRunner = require('../src/TestsRunner');
 
 const externalConfig = process.argv.slice(2);
-const ciNode = {
-  // nodes are indexed from 1 (not zero)
-  index: Number(process.env.CI_NODE_INDEX == null ? 1 : process.env.CI_NODE_INDEX),
-  total: Number(process.env.CI_NODE_TOTAL == null ? 1 : process.env.CI_NODE_TOTAL),
-};
```
tbergq, line 22: Not sure if it should be our primary concern anymore, but when I left kiwi, we were still relying on this for testing in east, west, and Europe timezone.

mrtnzlml: @tbergq I was thinking about this as well. There is an easy workaround to [execute the runner in parallel matrix](https://docs.gitlab.com/ee/ci/yaml/#parallel-matrix-jobs) (with different `TZ` env variable) if required. This simplifies our code and keeps exactly the same functionality.

My issue with keeping these company-specific parts is that people working in such companies are having a small interest in it which makes it impossible to know what they want and what they don't. So, from our perspective, it's an unused code and we are developing it in a certain direction we want which might or might not be aligned with the vision of other quiet users.

That being said, I'd be happy to keep these parts _if_ anyone interested gets involved (be it code contribution, sponsoring, or any other form). Hope that makes sense. 😎

tbergq: Yeah, I agree. I just wanted to put it out there. Now we even have history here pointing to a fix.

tbergq (approved):

PR #1531, opened 2020-12-16
Use latest Relay 10.1.2

mrtnzlml:
I am unable to run RN with the same Relay version. Not sure why. Running master does the trick.


mrtnzlml:
See: https://github.com/facebook/relay/releases/tag/v10.1.2

src/ya-comiste-react-native/package.json
```diff
@@ -28,19 +28,19 @@
     "react": "16.13.1",
     "react-native": "0.63.4",
     "react-native-navigation": "^7.6.0",
-    "react-relay": "0.0.0-master-fa1c75d2",
-    "relay-runtime": "0.0.0-master-fa1c75d2"
+    "react-relay": "0.0.0-master-c02d9722",
```
tbergq, line 31: Would it be possible to use the experimental here maybe? We could start playing with hooks?

tbergq (approved):

src/ya-comiste-react-native/package.json
mrtnzlml: @tbergq I think it would not because experimental Relay depends on unstable React which is not compatible with React Native requirements (it has pinned peer dependency on React 16.13.1 exactly).

PR #1540, opened 2020-12-19
SXT Plugin: Migrate to sxt attribute

michalsanger:
Breaking change in the use of SX Tailwind. Instead of `className={tailwind(...)}` only a `sxt="..."` in JSX is now used.


mrtnzlml (approved):
Cool change! 👍

PR #1545, opened 2020-12-22
SXT Plugin: Support conditional expression

michalsanger:


src/babel-plugin-transform-sx-tailwind/src/__tests__/__snapshots__/index.test.js.snap
```diff
@@ -100,6 +100,39 @@ const __styles_3xIUAp = sx.create({
...
+}
+
+      ↓ ↓ ↓ ↓ ↓ ↓
+
+// @flow
+import type { Element } from 'react';
+import sx from '@adeira/sx';
+export default function Example(isVisible: boolean): Element<'div'> {
+  return (
+    <div className={__styles_1spSrF(...(isVisible ? ['block'] : ['hidden']))}>Lorem lipsum</div>
```
mrtnzlml, line 122: It would be nice to have a fixture with multiple names in the array as well. Something like `sxt={isVisible ? 'aa bb cc' : 'dd ee ff'}` -> `className={__styles_1spSrF(...(isVisible ? ['aa', 'bb', 'cc'] : ['dd', 'ee', 'ff']))}`

mrtnzlml (approved):

PR #1547, opened 2020-12-22
YC: implement sessions management (auth/deauth)

mrtnzlml:
This commit implements `authorize_mobile` and `deauthorize_mobile` mutations as well as `whoami` query and request verification. This way, we can login on mobile via Google Sign-In, exchange the token on server for our session token and use it for the following requests. Eventually, we can easily implement more providers.

TODO:

- [x] manually test it


src/ya-comiste-meta/schema.graphql
```diff
@@ -35,6 +57,25 @@ type SDUISection {
 
 union SDUIComponent = SDUICardComponent | SDUIDescriptionComponent | SDUIJumbotronComponent | SDUIScrollViewHorizontalComponent
 
+type WhoamiPayload {
+  id: String
```
tbergq, line 61: Why is it String and not ID?

src/ya-comiste-meta/schema.graphql
```diff
@@ -26,6 +46,8 @@ type SDUICardComponent {
 
 type Query {
   mobileEntrypointSections(key: String!): [SDUISection!]!
+  "Returns information about the current user (can be authenticated or anonymous)."
+  whoami: WhoamiPayload!
```
tbergq, line 50: Why not use a union type?

src/ya-comiste-meta/schema.graphql
mrtnzlml: Yup, can be - changed.

src/ya-comiste-meta/schema.graphql
mrtnzlml: Because I find it unnecessarily complicated. It's there just for debugging purposes and it doesn't have any other real usage which would require better structure.

tbergq (approved):

PR #1576, opened 2020-12-31
CI: upload Rust build artifacts

mrtnzlml:
See: https://docs.github.com/en/free-pro-team@latest/actions/guides/storing-workflow-data-as-artifacts


mrtnzlml:
🤔 the uploaded artifact is missing the binary file

PR #1604, opened 2021-01-12
CI Rust: remove release build

mrtnzlml:
It takes a long time and it doesn't seem to be necessary (we are not using it for anything).


mrtnzlml:
@tbergq Could you please have a look at this one as well? :)

tbergq (approved):

PR #1621, opened 2021-01-15
YCR: add product search capabilities

mrtnzlml:


src/ya-comiste-rust/server/src/commerce/model/products.rs
```diff
@@ -122,26 +253,36 @@ async fn create_product_authorized(
...
     let insert_aql = arangors::AqlQuery::builder()
         .query(
             r#"
             INSERT {
-              name: @product_name,
-              description: @product_description,
               images: @product_images,
-              unit_label: "piece",
-              active: true,
+              unit_label: "product_units/1000844",
```
mrtnzlml, line 263: TODO: change this to the new value from migrations

PR #1633, opened 2021-01-18
CI Rust: always use colors for tests

mrtnzlml:


tbergq (approved):

mrtnzlml:
🤔 Hmm, it's still a bit strange because it doesn't color the tests output like on localhost. Let me check it again.

mrtnzlml:
I left there just the bare minimum because the colors of the actual tests didn't work. I don't know why (it should: https://github.com/rust-lang/cargo/issues/1983).

PR #1652, opened 2021-01-21
Open source new '@adeira/sx-design'

mrtnzlml:
The purpose of this package is to share common components written using SX (so it fits well with applications written is SX).

### DIFF
--- a/src/sx-design/src/.eslintrc.js
+++ b/src/sx-design/src/.eslintrc.js
@@ -1,0 +1,27 @@
+// @flow strict
+
+/* eslint-disable no-unused-vars */
+const OFF = 0;
+const WARN = 1;
+const ERROR = 2;
+/* eslint-enable no-unused-vars */
+
+module.exports = {
+  rules: {
+    // TODO: this should be true for all users of `@adeira/sx-design` so we should make it reusable?
+    'react/forbid-elements': [
+      'error',
+      {
+        forbid: [
+          { element: 'h1', message: 'use <Heading> instead' },
+          { element: 'h2', message: 'use <Heading> instead' },
+          { element: 'h3', message: 'use <Heading> instead' },
+          { element: 'h4', message: 'use <Heading> instead' },
+          { element: 'h5', message: 'use <Heading> instead' },
+          { element: 'h6', message: 'use <Heading> instead' },
+          { element: 'section', message: 'use <Section> instead' },
+        ],
+      },
+    ],
+  },
+};
--- a/src/sx-design/src/ProductCard.js
+++ b/src/sx-design/src/ProductCard.js
@@ -1,0 +1,34 @@
+// @flow
+
+import * as React from 'react';
+import sx from '@adeira/sx';
+
+import Heading from './Heading';
+
+export default function ProductCard(): React.Node {
+  // TODO: use Blurhash when loading the images
+
+  return (
+    <div className={styles('wrapper')}>
+      <Heading xstyle={styles.productTitle}>
+        <span>Product Name</span>
+      </Heading>
+    </div>
+  );
+}
+
+const styles = sx.create({
+  wrapper: {
+    height: 250,
+    width: 250,
+    display: 'flex',
+  },
+  productTitle: {
+    display: 'flex',
+    padding: 20,
+    color: '#1c1e21',
+    transitionProperty: 'background-color',
+    transitionDuration: '500ms',
+    transitionTimingFunction: 'ease-in-out',
+  },
+});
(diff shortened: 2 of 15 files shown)


src/sx-design/src/.eslintrc.js
```diff
@@ -0,0 +1,28 @@
...
+      {
+        forbid: [
+          { element: 'h1', message: 'use <Heading> instead' },
+          { element: 'h2', message: 'use <Heading> instead' },
+          { element: 'h3', message: 'use <Heading> instead' },
+          { element: 'h4', message: 'use <Heading> instead' },
+          { element: 'h5', message: 'use <Heading> instead' },
+          { element: 'h6', message: 'use <Heading> instead' },
+          { element: 'input', message: 'use <Input> or <InputSubmit> instead' },
+          { element: 'section', message: 'use <Seading> instead' },
```
tbergq, line 23: Should it be `use <Section> instead?`

mrtnzlml: It should. WTF. 😂

src/sx-design/src/ProductCard.js
```diff
@@ -0,0 +1,34 @@
...
+    height: 250,
+    width: 250,
+    display: 'flex',
+  },
+  productTitle: {
+    display: 'flex',
+    padding: 20,
+    color: 'var(--font-color-dark)',
+    transitionProperty: 'background-color',
+    transitionDuration: '500ms',
```
tbergq, line 31: This could potentially be a css variable too. 

Could be `--transition-duration-normal` , `--transition-duration-fast` etc.

src/sx-design/src/__tests__/Heading.test.js
```diff
@@ -0,0 +1,79 @@
...
+import Heading from '../Heading';
+import Section from '../Section';
+
+// https://reactjs.org/docs/testing-recipes.html
+
+let container: HTMLDivElement | null = null;
+beforeEach(() => {
+  // setup a DOM element as a render target
+  container = document.createElement('div');
+  document.body?.appendChild(container);
```
tbergq, line 18: why not use testing-library/react instead, that does all this setup and teardown for you?

src/sx-design/src/__tests__/Heading.test.js
```diff
@@ -0,0 +1,79 @@
...
+it('renders top level by default', () => {
+  // $FlowExpectedError[incompatible-call]
+  render(<Heading>level one</Heading>, container);
+  expect(container?.innerHTML).toMatchInlineSnapshot(`"<h1>level one</h1>"`);
+});
+
+// TODO: How should this actually behave? 🤔
+// <>
+//   <Heading>level one</Heading>   -> h1
+//   <Heading>level one</Heading>   -> h2
```
tbergq, line 37: Though it is normal to have only one h1 tag, I don't think it hurts your CEO score, nor screen readers having multiple ones h1's

tbergq:
What about setting up a storybook, so we could see the actual components?

src/sx-design/src/SkipLink.js
```diff
@@ -0,0 +1,32 @@
+// @flow
+
+import * as React from 'react';
+import sx from '@adeira/sx';
+
+type Props = {|
+  +text: FbtWithoutString,
+|};
+
+// https://web.dev/headings-and-landmarks/#bypass-repetitive-content-with-skip-links
```
tbergq, line 10: Seems to be some issues with this skip-link approach on mobile devices

https://axesslab.com/skip-links/

src/sx-design/src/__tests__/Heading.test.js
mrtnzlml: True. I captured the current behavior with a snapshot.

src/sx-design/src/SkipLink.js
mrtnzlml: Do you want to send a PR with some fix? It's not clear to me what should it look like from the article.

src/sx-design/src/ProductCard.js
mrtnzlml: It's actually an error that `color` was using a CSS var (copy-paste mistake). SX has a different way how to achieve customizability which I'd say is the way to go for now: https://github.com/adeira/universe/tree/master/src/sx#composability-and-customizability

mrtnzlml:
@tbergq Could you please have a look again?

src/sx-design/src/SkipLink.js
tbergq: Looks like [this](https://github.com/selfthinker/dokuwiki_template_writr/blob/master/js/skip-link-focus-fix.js) is the best solution. 

I can try to have a look at it later, should be ok merging this now, I just wanted to point out that it might not work on all devices 🙈

tbergq (approved):

PR #1655, opened 2021-01-22
Bazel: add experimental Rust support

mrtnzlml:


mrtnzlml:
Ugh, it takes a long time to complete. However, yesterday, I was playing around with https://docs.bazel.build/versions/master/remote-caching.html#google-cloud-storage and it looks quite promising.

tbergq:
> Ugh, it takes a long time to complete. However, yesterday, I was playing around with https://docs.bazel.build/versions/master/remote-caching.html#google-cloud-storage and it looks quite promising.

Nice, I also wanted to try this out 😊

mrtnzlml:
@tbergq The only thing which is bothering me a little bit is that I don't know how to set up it in GitHub because it requires a JSON file as an argument. 🤔

tbergq:
> @tbergq The only thing which is bothering me a little bit is that I don't know how to set up it in GitHub because it requires a JSON file as an argument. 🤔

What about keeping the content of the JSON file as a secret in GitHub, and write the JSON file on the CI?

There might be better solutions though, just the first thing that comes to mind 🤷‍♂️

mrtnzlml:
@tbergq That could work, yes. I'd like to check it closer though since it should work with S3 as well in which case I'd not expect `--google_credentials=…`.

mrtnzlml:
https://github.com/adeira/universe/pull/1672

mrtnzlml:
@tbergq So the remote cache is a significant boost. Actually, I'd say the slowest part is our shitty Bazel JavaScript implementation and the Rust part in Bazel is faster than running Rust specific things outside of Bazel. 🏃‍♂️ 

Do you think we could consider merging it to see how it behaves on a regular basis?

tbergq:
>Do you think we could consider merging it to see how it behaves on a regular basis?

Sure, yes, let's do so 👍

mrtnzlml:
Interesting discussion on this topic: https://github.com/google/cargo-raze/discussions/327 (+ some suggestions how could the structure and naming be improved/changed).

tbergq (approved):

mrtnzlml:
I am about to merge this, however, I already know about 2 issues:

- I do not know how to keep it in sync (ie. not to forget running `cargo raze` after deps upgrade)
- It's not for free (especially Google Cloud Storage egress)

Let's give it a try and let's see if it's viable.

PR #1702, opened 2021-02-01
NPM Publisher: Proper workflow failure status

### LINKED ISSUE #1671: NPM publisher stopped working
Something changed and our NPM publisher stopped working, see: https://github.com/adeira/universe/actions?query=workflow%3A%22NPM+Publisher%22

Also, seems like `npm-registry-client` (which we use for the publishing) is not maintained anymore.

michalsanger:
Let the workflow die correctly in case of NPM API error

Fix #1671

### DIFF
--- a/src/monorepo-npm-publisher/src/index.js
+++ b/src/monorepo-npm-publisher/src/index.js
@@ -110,6 +110,7 @@ export default async function publish(options: Options) {
           }
         }
       } catch (e) {
+        process.exitCode = 1;
         log('❌ Skipping %s because of NPM error: %s', chalkPackageName, e.message);
       }
     }


mrtnzlml:
> Fix #1671

@michalsanger Does this PR actually fix the issue? It looks like there is some other problem and we are unable to publish the package, or not? 🤔

michalsanger:
@mrtnzlml it fix the issue when other packages were not published when there was a NPM API issue with one of them. 

It does not fix this:
> Skipping @adeira/sx-design because of NPM error: 403 Forbidden

mrtnzlml:
@michalsanger So I'd suggest not to close the original issue, what do you think?

mrtnzlml (approved):

PR #1707, opened 2021-02-02
Bazel: remove Rust support

mrtnzlml:
The only reason why I want to do this is the cost: we are having approx. 10 gibibyte egress per day with GCP and Bazel Remote Cache which equals ~1 USD per day. In other words, this experiment costs 30 USD per month which is simply too much for non-profit project. Especially considering we don't need it.

However, it's definitely viable and way to go. It's just too early.

TODO when approved:

- [ ] remote GCP credentials from GitHub
- [ ] remove GCP system account


mrtnzlml:
@adeira/devs Could you approve, please?

michalsanger (approved):

tbergq (approved):

Issue #1718, opened 2021-02-05
Deprecate Node.js 12


mrtnzlml:
_Based on this discussion: https://github.com/adeira/universe/discussions/1588_

The following steps should be applied in this exact order:

- [x] (https://github.com/adeira/universe/pull/1720) write the following paragraph into `CHANGELOG.md` files of **all packages** (please help me with the wording):

> Support for Node.js 12 has been removed. This package might continue working on older Node.js versions, however, it's highly recommended to upgrade to Node.js version 14 or newer. For more details, see: https://nodejs.org/en/about/releases/

- [x] major bump all packages with version >=1.0.0 (might be unnecessary in some cases):
    - [x] `@adeira/babel-preset-adeira` (https://github.com/adeira/universe/pull/1748)
    - [x] `@adeira/css-colors` (https://github.com/adeira/universe/pull/1729)
    - [x] `@adeira/eslint-config` (https://github.com/adeira/universe/pull/1735)
    - [x] `@adeira/eslint-runner` (https://github.com/adeira/universe/pull/1744)
    - [x] `@adeira/fetch` (https://github.com/adeira/universe/pull/1747)
    - [x] `@adeira/graphql-global-id` (https://github.com/adeira/universe/pull/1745)
    - [x] `@adeira/js` (https://github.com/adeira/universe/pull/1746)
    - [x] `@adeira/monorepo-npm-publisher` (https://github.com/adeira/universe/pull/1743)
    - [x] `@adeira/murmur-hash` (https://github.com/adeira/universe/pull/1730)
    - [x] `@adeira/relay` (https://github.com/adeira/universe/pull/1749)
    - [x] `@adeira/signed-source` (https://github.com/adeira/universe/pull/1731)

- [x] (https://github.com/adeira/universe/pull/1750) remove Node.js 12 from CI (once every package with version >= 1.0.0 is released)
- [x] (https://github.com/adeira/universe/pull/1751) _optional_ update code in `eslint-plugin-sx` which would greatly benefit from removing Node.js 12 limitations

michalsanger:
> remove Node.js from CI (once every package with version >= 1.0.0 is released)

You mean "remove Node.js v12 from CI"?

PR #1720, opened 2021-02-05
Changelogs: add Node.js 12 deprecation notice

mrtnzlml:
See: https://github.com/adeira/universe/issues/1718


mrtnzlml:
Tests fix here: https://github.com/adeira/universe/pull/1727

tbergq (approved):

PR #1760, opened 2021-02-11
upgrade babel packages

tbergq:


src/example-relay/.babelrc.js
```diff
@@ -2,5 +2,5 @@
 
 module.exports = {
   presets: ['@adeira/babel-preset-adeira', 'next/babel'],
-  plugins: ['relay'],
+  plugins: ['relay', '@babel/plugin-proposal-class-properties'],
```
tbergq, line 5: This fixes the failing test-bc script locally, but I have no idea why we have to add it since it should already be in our babel-prest

mrtnzlml: 🤔

mrtnzlml (approved):

PR #1763, opened 2021-02-12
YCR: allow multilingual create product input

mrtnzlml:
This way we can create the product in spanish, english, or both at the same time.

### DIFF
--- a/src/ya-comiste-rust/server/src/commerce/dal/products.rs
+++ b/src/ya-comiste-rust/server/src/commerce/dal/products.rs
@@ -1,0 +1,297 @@
+use crate::arangodb::ConnectionPool;
+use crate::commerce::model::errors::ModelError;
+use crate::commerce::model::products::{
+    ClientLocale, PriceSortDirection, Product, ProductMultilingualInput,
+};
+use serde_json::json;
+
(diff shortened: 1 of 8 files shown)


mrtnzlml:
Leaving it here for tomorrow when I am less tired.

src/ya-comiste-rust/server/src/commerce/dal/products.rs
```diff
@@ -0,0 +1,297 @@
...
+              translations: MERGE(translations)
+            } INTO products
+            LET product = NEW
+            RETURN UNSET(
+              MERGE_RECURSIVE(product, product.translations[@product_language]),
+              "translations"
+            )
+            "#,
+        )
+        .bind_var("product_language", String::from("en_US")) // TODO
```
mrtnzlml, line 53: 🤔 This is good for fetching on FE but what if BO needs to fetch both language versions at the same time? (for product editation for example)

mrtnzlml: I should probably add a new field `translations` to the output because it makes sense to return the title in the locale we have BO/FE but also return multiple translation versions if needed (for the edits for example).

PR #1805, opened 2021-02-19
Remove 'next/babel' preset

mrtnzlml:
This partially reverts 1c5d0b46ea5f0814a6d3b3f1566c2b7d8d68e908

Something changed somewhere somehow. I faced similar issue yesterday with broken class properties plugin and Flow types so I had to do something. This seems to be fixing the issue at least for now.

### DIFF
--- a/src/example-relay/.babelrc.js
+++ b/src/example-relay/.babelrc.js
@@ -1,13 +1,6 @@
 // @flow strict
 
 module.exports = {
-  presets: ['@adeira/babel-preset-adeira', 'next/babel'],
-  /**
-   * test-bc script broke after upgrading some babel packages.
-   * Adding @babel/plugin-proposal-class-properties made it work, but it should not be necessary
-   * to add it here, since we already have it in @adeira/babel-preset-adeira.
-   * We need to leave it here for now for the test-bc script to work, but we should try
-   * to remove it in the future
-   */
-  plugins: ['relay', '@babel/plugin-proposal-class-properties'],
+  presets: ['@adeira/babel-preset-adeira'],
+  plugins: ['relay'],
 };
--- a/src/example-relay/__github__/babel.config.js
+++ b/src/example-relay/__github__/babel.config.js
@@ -21,7 +21,7 @@ module.exports = function (api /*: ApiType */) /*: BabelConfig */ {
   api.cache.forever();
 
   return {
-    presets: ['@adeira/babel-preset-adeira', 'next/babel'],
+    presets: ['@adeira/babel-preset-adeira'],
     plugins: ['relay'],
   };
 };
--- a/src/sx-tailwind-website/.babelrc.js
+++ b/src/sx-tailwind-website/.babelrc.js
@@ -4,7 +4,7 @@ const defaultTheme = require('tailwindcss/defaultTheme');
 const colors = require('tailwindcss/colors');
 
 module.exports = {
-  presets: ['@adeira/babel-preset-adeira', 'next/babel'],
+  presets: ['@adeira/babel-preset-adeira'],
   plugins: [
     [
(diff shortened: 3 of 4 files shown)


mrtnzlml:
cc @tbergq

mrtnzlml:
And now the `scripts/test-bc.js` file doesn't work again for whatever reason. 🤦‍♂️

tbergq:
> And now the `scripts/test-bc.js` file doesn't work again for whatever reason. 🤦‍♂️

Yeah, there is a little gotcha, we are not running this script locally with our `yarn test`, we should probably add it there.

tbergq:
Maybe we should have a separate babel-preset for our next-apps, it seems like our preset is not playing nicely with `next/babel`?

mrtnzlml:
> Yeah, there is a little gotcha, we are not running this script locally with our yarn test, we should probably add it there.

💡 Aha, that's why it's only when running the script directly. I thought it's somehow related to running tests vs. running monorepo-babel-node.

> Maybe we should have a separate babel-preset for our next-apps, it seems like our preset is not playing nicely with `next/babel`?

But why doesn't it play nicely? I don't understand. I am not sure if our preset is the one to blame. 🤔

tbergq:
Well, it just feels like if the order matters and we have to add additional plugins that are in the preset because we use these 2 presets together, they don't work well together.

I am not sure ours is to blame either. The important part is that we can build or app and run the test scripts 😊

mrtnzlml:
@tbergq OK, so I'd like to propose a removal of `next/babel` preset altogether. Please have a look at this updated PR.

tbergq (approved):

tbergq:
I'll approve it, but it would be nice to double check what it actually adds, and verify that we are not missing some plug-in. 

Also, I see that the pipeline is still failing.

mrtnzlml:
> I'll approve it, but it would be nice to double check what it actually adds, and verify that we are not missing some plug-in.

We are definitely missing some Next.js specifics, see: https://github.com/vercel/next.js/blob/4b587d34b867b4f5bad20d520b789f1543249df2/packages/next/build/babel/preset.ts

I don't know how to push it forward though. Having `@babel/plugin-proposal-class-properties` in plugins is also a bad solution because we have to eventually start using [`declare` fields](https://babeljs.io/docs/en/babel-plugin-transform-flow-strip-types#allowdeclarefields) (before Babel 8) and this made it impossible. 🤔 It starts throwing the following error:

```text
SyntaxError: /Users/martinzlamal/Work/adeira/universe/src/example-relay/src/graphql/locations/types/output/AddLocationOrError.js: TypeScript 'declare' fields must first be transformed by @babel/plugin-transform-typescript.
If you have already enabled that plugin (or '@babel/preset-typescript'), make sure that it runs before any plugin related to additional class features:
 - @babel/plugin-proposal-class-properties
 - @babel/plugin-proposal-private-methods
 - @babel/plugin-proposal-decorators
   8 |
   9 | export class ValidLocationResponse {
> 10 |   declare location: $ReadOnly<{| ...LocationInput, +id: string |}>;
     |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  11 |   constructor(location: LocationInput) {
  12 |     this.location = {
  13 |       ...location,
```

Notice the Typescript-specific errors, WTF.

> Also, I see that the pipeline is still failing.

I think it's failing because of this which is unrelated: https://github.com/adeira/universe/issues/1825

mrtnzlml:
I am closing this for now because I am not confident about what's actually going on. 😞 

#help_needed 😬

mrtnzlml:
https://github.com/adeira/universe/issues/1854

PR #1809, opened 2021-02-19
SX: implement conditional object style selector

mrtnzlml:
Previously, you could select styles conditionally only using the following syntax:

```
styles('button', isDisabled && 'disabled');
styles('button', isDisabled ? 'disabled' : null);
```

Now it's possible to use an object for these conditionals:

```
styles({ button: true, disabled: isDisabled });
```

This is especially useful when all the styles are conditional. For example, imagine a case where you can change the component style with one React property:

```
styles({
  primary: props.color === 'primary',
  highlight: props.highlight === 'highlight',
});
```


mrtnzlml:
Resolves another point from this list: https://github.com/adeira/universe/discussions/996

tbergq (approved):

PR #1822, opened 2021-02-22
Relay: Fetch schema script options fix

### LINKED ISSUE #1814: @adeira/relay: unable to use 'adeira-fetch-schema' script
To reproduce:

```bash
yarn adeira-fetch-schema --resource …
```

Throws:

```text
Invariant Violation: Option --resource is required.
```

michalsanger:
Fix #1814

### DIFF
--- a/src/relay/CHANGELOG.md
+++ b/src/relay/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fetch schema script fixed
+
 # 3.0.0
 
 - Support for Node.js 12 has been removed. This package might continue working on older Node.js versions, however, it's highly recommended upgrading to Node.js version 14 or newer. For more details, see: https://nodejs.org/en/about/releases/, or discuss here: https://github.com/adeira/universe/discussions/1588
--- a/src/relay/bin/fetch-schema.js
+++ b/src/relay/bin/fetch-schema.js
@@ -24,12 +24,14 @@ program
   .option('--filename <path>', undefined, 'schema.graphql')
   .parse(process.argv);
 
-invariant(program.resource, 'Option --resource is required.');
+const options = program.opts();
 
-const filename = path.join(process.cwd(), program.filename);
+invariant(options.resource, 'Option --resource is required.');
+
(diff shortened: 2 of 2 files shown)


mrtnzlml (approved):

mrtnzlml:
I am wondering if there is an easy way how to test this? 🤔 It probably broke with some dependency breaking change upgrade.

michalsanger:
I wish I had an idea how to test it :-(

mrtnzlml:
OK, no problemo. Thanks for the fix.

tbergq:
> I am wondering if there is an easy way how to test this? 🤔 It probably broke with some dependency breaking change upgrade.

I had an idea about doing something like: 

```js
spawnSync(
    'node',
    [
      require.resolve('../fetch-schema'),
      `--filename=${path.relative(
        findMonorepoRoot(),
        require.resolve('../__fixtures__/test.graphql'),
      )}`,
      `--resource=https://www.test.gql`,
    ],
    { stdio: 'inherit', cwd: findMonorepoRoot() },
  );
```

and use `msw` or mock fetch and read the file it writes and assert it as a snapshot. 

```js
expect(
    fs.readFileSync(path.join(__dirname, '..', '__fixtures__', 'test.graphql'), 'utf-8'),
  ).toMatchSnapshot();
```

but that didn't work, since I spawn a new process and neither msw or the mocked fetch exists in that new process.

tbergq:
Actually, I might have found a way to make it work. I will see if I can clean it up tomorrow.

mrtnzlml:
Such a high-level integration test might not be necessary. After all, the issue was that we were using `commander` incorrectly, and could not retrieve the CLI options. It would be enough to do something like this to prevent this issue in the future: https://github.com/tj/commander.js/blob/4aaaa9de3fd09401afe3894483193ba0e2e512d7/tests/args.literal.test.js#L9-L18 (so simply extract the options definition and try to parse some fake command, no need to run the whole fetcher). The second follow-up of this PR should be to install Flow types for `commander` from `flow-typed`. Does anyone want to do it? 😊

PR #1826, opened 2021-02-22
Upgrade '@babel/runtime' to the latest version 7.12.18

mrtnzlml:
See: https://github.com/babel/babel/blob/ba4ac7d58738b94425026e1599d95e6bf9b2ba68/CHANGELOG.md#v71218-2021-02-18


mrtnzlml:
Dependabot is timeouting so some manual work is needed.

PR #1855, opened 2021-02-23
Babel preset: turn on 'allowDeclareFields' for Flow code

mrtnzlml:
In preparation for Babel 8. Basically, there was a bug in Babel which was incorrectly stripping type only fields resulting in invalid code. To mitigate this issue, Flow added special `declare` field and Babel will be stripping only these fields (while leaving the uninitialized fields as undefined).

See:

- https://babeljs.io/docs/en/babel-plugin-transform-flow-strip-types#allowdeclarefields
- https://github.com/babel/babel/pull/11178
- https://github.com/facebook/flow/commit/11b7adbf85ea237cfc01d1b8dc6f6dcdbc299157


src/adeira.dev/package.json
```diff
@@ -11,7 +11,7 @@
     "write-translations": "docusaurus write-translations"
   },
   "dependencies": {
-    "@docusaurus/core": "^2.0.0-alpha.69",
+    "@docusaurus/core": "^2.0.0-alpha.70",
```
mrtnzlml, line 14: Unrelated but should be like this. 😬

michalsanger (approved):

PR #1890, opened 2021-03-02
Merge Relay packages together

mrtnzlml:
Previously, we tried to split the packages similarly to how Relay does it internally. However, we never finished this effort leaving it fragmented. I'd like to propose merging it back together with these steps to follow:

- cleanup the package (remove things we don't need to wrap because Relay already provides the necessary Flow types)
- get it ready for next major version Relay 11


scripts/publishedPackages.json
```diff
@@ -20,8 +20,6 @@
   "@adeira/monorepo-utils",
   "@adeira/murmur-hash",
   "@adeira/relay",
-  "@adeira/relay-runtime",
-  "@adeira/relay-utils",
```
tbergq, line 24: Maybe we should publish a new version with these packages where we update the readme and say that they are deprecated. Similar to [react-axe](https://www.npmjs.com/package/react-axe)

mrtnzlml: I am going to deprecate them directly on NPM (so it's well-integrated and you should see it when installing). Additionally, these packages never reached a stable state so should not be a big deal.

tbergq (approved):

michalsanger (approved):

PR #1893, opened 2021-03-02
Relay: remove unnecessary `commitLocalUpdate` wrapper

mrtnzlml:
It's not needed anymore.


tbergq (approved):
Seems like you need to update a snapshot though.

PR #1900, opened 2021-03-03
Relay: add tests for Commander CLI options

mrtnzlml:
Followup after: de8ed188a82047aa25237f5e279135652e963e2e


src/relay/bin/commander/__tests__/options.test.js
```diff
@@ -0,0 +1,106 @@
...
+    relayCompilerOptions = require('../options').relayCompilerOptions;
+  });
+});
+
+describe('fetch schema', () => {
+  it('no args', () => {
+    expect(fetchSchemaOptions(['node', 'test'])).toMatchInlineSnapshot(`
+      Object {
+        "filename": "schema.graphql",
+      }
```
tbergq, line 17: Why use snapshot here instead of just `toEqual({filename: 'schema.graphql'})`?

tbergq (approved):

src/relay/bin/commander/__tests__/options.test.js
mrtnzlml: Because I am often looking for lazy solutions but using the equality tests is obviously better. Please have a look again.

tbergq (approved):

PR #1932, opened 2021-03-08
SX Design: add `ErrorBoundary` component

mrtnzlml:
Not only this adds a new `ErrorBoundary` component, but I also tried to kickoff experimental FBT support. I still have to figure out how to integrate the localization/translations properly so it's not awkward.


src/sx-design/src/ErrorBoundary/ErrorBoundary.js
```diff
@@ -0,0 +1,97 @@
...
+  render(): Node {
+    if (this.state.hasError) {
+      return (
+        <div className={styles('error')}>
+          <div>
+            <Heading xstyle={styles.h1}>{this.props.code ?? '5XX'}</Heading>
+            <Section xstyle={styles.desc}>
+              <Heading xstyle={styles.h2}>
+                {this.props.title ??
+                  fbt('An unexpected error has occurred.', 'generic error message')}
```
tbergq, line 48: What about also displaying a retry button, and allow the consumer to pass in an `onRetry` callback? by default it could do `location.reload` maybe?

tbergq (approved):

PR #1947, opened 2021-03-09
Use Relay version 11.0.0

mrtnzlml:
See: https://github.com/facebook/relay/releases/tag/v11.0.0


michalsanger (approved):

mrtnzlml:
Just a note: I'd like to release `@adeira/relay` as soon as possible. Would be nice to convert all our projects first but it's not a requirement. The next version of `@adeira/relay` should be minor (not breaking) so everyone can easily upgrade step by step.

PR #1961, opened 2021-03-11
Flow: use implicit exact object types

mrtnzlml:
Flow standard is now to use `{ }` for strict objects and `{ key: value, ... }` for open objects, see: https://github.com/facebook/relay/commit/6fa0b0d83b45af67d086e5b850b4b49915eccb4a and https://github.com/facebook/react-native/commit/00cfb0f919207fcdc8fe8b6c3b7d76591447b91d

This change should be without any problems as longs as users of our NPM libraries are using Flow with `exact_by_default=true` (which is now pretty standard I'd say).

Here is how object types behave before this change:

```
type A = { x: number }; // already "exact" type but disallowed because it would be confusing to mix different syntaxes
type B = { x: number, ... }; // this is inexact
type C = {| x: number |}; // this is explicitly exact and the only allowed way how to describe type "exactness"
```

Here is how object types behave _after_ this change:

```
type A = { x: number }; // this is the only allowed syntax for "exact" type
type B = { x: number, ... }; // this is still inexact
type C = {| x: number |}; // this is also exact but no longer allowed (so it's not confusing)
```

Some related (non-blocking) issues:

- https://github.com/gajus/eslint-plugin-flowtype/issues/467
- https://github.com/facebook/flow/issues/8612

TODO:

- [x] [backward compatibility: what about exported NPM libs with these new types](https://github.com/mrtnzlml/test-flow-exact-by-default-compatibility)
- [x] coordinate with next major Eslint Config release: https://github.com/adeira/universe/pull/2159

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,4 +1,4 @@
-# @generated SignedSource<<ae49cb9ae55943d5887942a3c1b7b525>>
+# @generated SignedSource<<cc03860d3a8d355bd35e54f6ada88c23>>
 #
 # To regenerate run:
 # yarn monorepo-babel-node scripts/generateFlow.js
@@ -24,10 +24,10 @@
 
 [lints]
 all=error
-ambiguous-object-type=off
 sketchy-null-bool=off
 unclear-type=off
 untyped-import=off
+ambiguous-object-type=off
 
 [options]
 emoji=true
@@ -42,7 +42,6 @@ experimental.const_params=true
 [rollouts]
 
 [strict]
-ambiguous-object-type
 sketchy-null-bool
 unclear-type
 untyped-import
--- a/.flowconfig.template.base
+++ b/.flowconfig.template.base
@@ -35,10 +35,13 @@
 ; https://flow.org/en/docs/config/lints/
 [lints]
 all=error
-ambiguous-object-type=off
 sketchy-null-bool=off
 unclear-type=off
 untyped-import=off
+
+# Flow standard is now to use `{ }` for strict objects and `{ key: value, ... }` for open objects,
+# so this option should always be `off`.
+ambiguous-object-type=off
 
 
 ; This setting complements [lints] section: all the enabled lints still throw errors + these
@@ -47,7 +50,6 @@ untyped-import=off
 ;
 ; https://flow.org/en/docs/strict/
 [strict]
-ambiguous-object-type
 sketchy-null-bool
 unclear-type
 untyped-import
--- a/src/__flowtests__/ReactElement.js
+++ b/src/__flowtests__/ReactElement.js
@@ -4,30 +4,30 @@
 
 import { Component, type Element, type ChildrenArray, type Node } from 'react';
 
-class Button extends Component<{| +disabled?: boolean |}> {
+class Button extends Component<{ +disabled?: boolean }> {
   render() {
     return null;
   }
 }
 
-class DisabledButton extends Component<{||}> {
+class DisabledButton extends Component<{}> {
   // The return type is not necessary - it's here only to demonstrate what is going on.
   render(): Element<typeof Button> {
     return <Button disabled={true} />;
   }
 }
 
-class WrapperLimited extends Component<{|
+class WrapperLimited extends Component<{
   +children: ChildrenArray<
     // You have to specify every single supported component here.
     Element<typeof Button> | Element<typeof DisabledButton>,
   >,
-|}> {}
+}> {}
 
-class WrapperSmart extends Component<{|
+class WrapperSmart extends Component<{
   // Type `RestrictedElement` understands what is being rendered so it accepts even `DisabledButton`(because it returns `Button`).
   +children: ChildrenArray<RestrictedElement<typeof Button>>,
-|}> {}
+}> {}
 
 module.exports.testStupid = ((
   <WrapperLimited>
--- a/src/__flowtests__/RestrictedElement.js
+++ b/src/__flowtests__/RestrictedElement.js
@@ -5,14 +5,14 @@
 import * as React from 'react';
 
 // You can enforce compositional patterns with RestrictedElement:
-class MenuItem extends React.Component<{||}> {}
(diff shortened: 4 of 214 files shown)


mrtnzlml:
Draft only for now: still investigating what gets broken, what are the consequences, and how to make sure we keep the codebase unified.

mrtnzlml:
@adeira/devs could you please have a look at this PR draft and share your opinion/concerns? Thanks! 😎

michalsanger:
I'd say let’s do this.

> think about backward compatibility

What exactly do you mean by this? BC within the Universe monorepo? Can it be achieved and do we even need it? 

> what about exported NPM libs with these new types?

Could a new major release be the solution?

mrtnzlml:
> > think about backward compatibility
> 
> What exactly do you mean by this? BC within the Universe monorepo? Can it be achieved and do we even need it?

For example, how does it behave when we publish an NPM package with this new setup and users of our package are using the old setup? I would like to know the answer before giving this PR for a proper review.

I guess both of these TODOs are the same and I had the question twice without noticing. 😬 

> > what about exported NPM libs with these new types?
> 
> Could a new major release be the solution?

It can be a solution, yes. I'd like to check first if it's necessary though. Also, Flow usage is quite low so we probably can cut some corners.

michalsanger:
If I understand this correctly, when you are using the old syntax and start mixing it with the new one, you are basically using weaker types. You don't get any Flow errors. In this meaning it's backward compatible, but from DX perspective I would appreciate a big info in changelog. Major version could be that big info, on the other hand do you see any libs with Flow types doing this?

mrtnzlml:
> If I understand this correctly, when you are using the old syntax and start mixing it with the new one, you are basically using weaker types. You don't get any Flow errors. In this meaning it's backward compatible, but from DX perspective I would appreciate a big info in changelog. Major version could be that big info, on the other hand do you see any libs with Flow types doing this?

I believe this is not correct. There are many changes in this PR but they are mostly syntactical - no big changes in terms of behavior. That's thanks to the fact that we are using "exact by default" behavior for a long time. Here is how it is now:

```js
export type A = { x: number }; // this is already "exact" type but disallowed because it would be confusing to mix different syntaxes
export type B = { x: number, ... }; // this is inexact
export type C = {| x: number |}; // this is explicitly exact and the only allowed way how to describe type "exactness"
```

Here is what this PR changes:

```js
export type A = { x: number }; // this is the only allowed syntax for "exact" type
export type B = { x: number, ... }; // this is still inexact
export type C = {| x: number |}; // this is also exact but no longer allowed (so it's not confusing)
```

That's why I believe that changing `{| x: number |}` to `{ x: number }` in our NPM libraries should be fine (with no BC break and no change in the type soundness) **as long as you use `exact_by_default=true`**. This feature was rolled out as follows:

- [v0.101.0](https://github.com/facebook/flow/releases/tag/v0.101.0) (10 Jun 2019) new `implicit-inexact-object` lint
- [v0.105.0](https://github.com/facebook/flow/releases/tag/v0.105.0) (9 Aug 2019) new `exact_by_default` option in flowconfig
- we started using `exact_by_default=true` 9 Nov 2019, see: https://github.com/adeira/universe/commit/a9037e62324e0b8290fa8b9918d5974934a4ecc9
- [v0.112.0](https://github.com/facebook/flow/releases/tag/v0.112.0) (12 Nov 2019) new `ambiguous-object-type` lint
- v0.148.0 is the latest Flow version

This timeline makes me think that it should be fine to proceed (6 Apr 2021). Please note though that I didn't actually try it yet so I might be wrong.

mrtnzlml:
We actually attempted to do this change more than one year ago but I guess it was too early: https://github.com/adeira/universe/pull/362

mrtnzlml:
Here is my testing: https://github.com/mrtnzlml/test-flow-exact-by-default-compatibility

TL;DR:

# Conclusion

## Users with `exact_by_default=false` (legacy default)

Change from explicit exact objects (`{||}`) to implicit exact objects (`{}`) can have unintended effects. Such users should convert their codebase as soon as possible:

- first to explicit exact objects (`{||}`) and explicit inexact objects (`{...}`)
- second from explicit exact objects (`{||}`) to implicit exact objects (`{}`) while keeping explicit _inexact_ objects

## Users with `exact_by_default=true` (the old new)

There is no change and everything should be good to go. ✅

mrtnzlml:
> …from DX perspective I would appreciate a big info in changelog. Major version could be that big info, on the other hand do you see any libs with Flow types doing this?

@michalsanger What would you suggest to write into the changelog? Seems like FB is [migrating mainly internal code](https://github.com/facebook/react-native/commit/00cfb0f919207fcdc8fe8b6c3b7d76591447b91d) and [docs](https://github.com/facebook/relay/commit/6fa0b0d83b45af67d086e5b850b4b49915eccb4a). I wouldn't personally release new major versions because of this change. Realistically, usage of Adeira libs is very small let alone their Flow version.

michalsanger:
OK, pretty clear now. Releasing a major version is overkill, standard info in the changelog should be enough.

michalsanger (approved):

mrtnzlml:
@tbergq Do you have some input as well?

mrtnzlml:
I am going to put the following note into _all_ changelog files:

> Note for all Flow users: all projects in [`adeira/universe`](https://github.com/adeira/universe) now use implicit exact Flow types (`{}` for strict objects and `{ ... }` for open objects, syntax `{||}` is deprecated). We do not expect any issues as long as you are using `exact_by_default=true` Flow option.

OK?

tbergq (approved):

mrtnzlml:
OSS versions of FBT and Recoil are now also getting this change, see: https://github.com/search?q=%22fb-www%2Fflow-exact-by-default-object-types%22&type=commits ([fbt](https://github.com/facebook/fbt/commit/80e2f116074d2db7f0e37bc238dce05025cb781e), [recoil](https://github.com/facebookexperimental/Recoil/commit/8426b708a8f7bdd2a5f814751d020833d98aeb0b))

PR #2028, opened 2021-03-22
SX Eslint: improve detection of unused styles for conditional objects

mrtnzlml:


src/eslint-plugin-sx/src/rules/__tests__/fixtures/no-unused-stylesheet/invalid/sx-conditional-call-2.js
```diff
@@ -0,0 +1,20 @@
...
+ * @eslintExpectedError (13:40;13:61) Unknown stylesheet used: ccc (not defined anywhere)
+ * @eslintExpectedError (19:3;19:29) Unused stylesheet: xxx (defined via "styles" variable)
+ * @flow
+ */
+
+import type { Node } from 'react';
+import { create as sxCreate } from '@adeira/sx';
+
+export default function MyComponent(): Node {
+  const isBBB = Boolean(Math.random());
```
michalsanger, line 11: I don't get this. You want a random boolean? `Math.random() < 0.5` would do the trick, but `const isBBB = false;` would work just fine in a context of this test no?

mrtnzlml: You are right that the code doesn't make much sense. I don't remember the history of it (now it's just a copy-paste thing) but I know that it was very convenient to have there some `CallExpression` other than `styles(…)` from SX because it was colliding and not working correctly.

So maybe something like `const isBBB = String("unrelated CallExpression")` would be better? It still doesn't make much more sense but maybe it's a bit more obvious?

mrtnzlml: Alternatively, I could create a separate fixture capturing the edge-case with CallExpression and simply use isBBB = false as you suggested. That would probably be the best. In a follow-up PR?

michalsanger: Yeah, edge-case with CallExpression sounds like a best option.

michalsanger (approved):

src/eslint-plugin-sx/src/rules/__tests__/fixtures/no-unused-stylesheet/invalid/sx-conditional-call-2.js
mrtnzlml: I updated it in this PR. Turned out there should already be fixtures covering the `CallExpression` cases.

PR #2075, opened 2021-03-28
Downgrade '@babel/core' dependency

mrtnzlml:
Some tests started failing with the following error:

```
Error while loading config - (0 , _helperCompilationTargets(...).isBrowsersQueryValid) is not a function
```

The failing seem to be quite stable on CI but locally the errors seem to be fairly random.

I found this issue which looks similar: https://github.com/babel/babel/issues/12732

So I tried to enforce only one `@babel/core` dependency (instead of having several conflicting versions) and all seem to be good now (at least locally).

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "@adeira/eslint-runner": "^2.0.0",
     "@adeira/flow-bin": "^0.8.0",
     "@babel/cli": "^7.13.10",
-    "@babel/core": "^7.13.10",
+    "@babel/core": "~7.12.17",
     "babel-eslint": "^10.1.0",
     "babel-jest": "^26.6.3",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -37,6 +37,6 @@
     "typecheck": "yarn run adeira-flow-bin"
   },
   "resolutions": {
-    "graphql": "^15.0.0"
```
mrtnzlml, line 40: GraphQL resolution is no longer needed. At least I think.

mrtnzlml:
OK, it doesn't look like it helped very much. Now it fails differently. 🤦

package.json
mrtnzlml: https://github.com/adeira/universe/pull/2081

PR #2110, opened 2021-03-31
Kodiak: Improved automerge setup

michalsanger:
This could make "automerge priority" label to work alone.

Currently you have to add "automerge" together with "automerge priority".

See https://kodiakhq.com/docs/config-reference#mergeautomerge_label


michalsanger:
Maybe it will work after it's in `master`

PR #2169, opened 2021-04-08
SX Design: update CSS variables

mrtnzlml:
I took a great inspiration from https://vercel.com/design/color. The main change (other than adding new variables for colors) is that I got rid of "text color" and replaced it with "foreground color". This allows use to better describe the actual use-case when there is no text involved.

I have other PRs in the queue using these colors so stay tuned.


src/sx-design/src/Heading/__tests__/__snapshots__/Heading.test.js.snap
```diff
@@ -1,30 +1,30 @@
...
-    <section class=\\"_4anm7e\\">
-      <h4 class=\\"_4anm7e\\">level four</h4>
-      <section class=\\"_4anm7e\\">
-        <h5 class=\\"_4anm7e\\">level five</h5>
-        <section class=\\"_4anm7e\\">
-          <h6 class=\\"_4anm7e\\">level six</h6>
-          <section class=\\"_4anm7e\\">
-            <h6 class=\\"_4anm7e\\">level six again</h6>
-            <h6 class=\\"_4anm7e\\">level six again</h6>
+"<h1 class=\\"_1cJrQ0\\">level one</h1>
```
mrtnzlml, line 4: This sucks but we do not have a better solution for now. It's described here as "Jest snapshot serializer": https://github.com/adeira/universe/discussions/996

michalsanger (approved):

PR #2212, opened 2021-04-13
Relay example: Use hooks for pagination

michalsanger:


tbergq (approved):

src/example-relay/src/Homepage/locations/LocationsPaginated.js
```diff
@@ -1,71 +1,60 @@
...
-} from '@adeira/relay';
+import { type Node, useCallback } from 'react';
+import { graphql, usePaginationFragment } from '@adeira/relay';
 
 import Button from '../../components/Button';
 import Location from './Location';
 import LocationList from './LocationsList';
-import type { LocationsPaginated_data as LocationsDataType } from './__generated__/LocationsPaginated_data.graphql';
+import { type LocationsPaginatedRefetchQuery } from './__generated__/LocationsPaginatedRefetchQuery.graphql';
+import { type LocationsPaginated_data$key as LocationsFragmentRef } from './__generated__/LocationsPaginated_data.graphql';
```
mrtnzlml, line 10: There is a cool way how to avoid aliasing the import (I assume because of `camelcase` lint rule?), _if you want_. Simply name the fragment the way you want it to be (without `*_data` postfix) and you are good to go. Using `<FileName>_<propName>` is an old convention that is no longer needed.

mrtnzlml:
There is an interesting bug (didn't have time to dig into it yet so just steps to reproduce):

1) click "Load more" under "Incremental pagination 2"
2) try to click "Load more" under "Incremental pagination 1" 💥 

In my case, the "Incremental pagination 1" doesn't do anything. BUT, if you do it in any other order, everything works fine. It's broken only when the last column is longer than the other one. 🤔 

Also, there is a failing test (`LocationsPaginatedRefetch.test.js`) that seems to be pointing to something similar. Maybe helpful for understanding what's going on.

michalsanger:
Not sure why, but `fetchPolicy: 'store-and-network'` helped to fix the bug. 

Also unclear why, but the failing test was fixed when the `environment.mock.resolveMostRecentOperation` was moved into `act(...)`

mrtnzlml (approved):
Well done!

mrtnzlml:
@michalsanger Could you please update README as well? Can be in a different PR.

PR #2270, opened 2021-04-21
SX Design: kick off RTL support

mrtnzlml:
LTR and RTL layouts are another important aspect of good design systems. I want to kick off this effort even though I do not have a real use case for RTL languages. I believe that applying additional constraints on this design system (and therefore SX itself) will help us make it more flexible and ready for more complex scenarios.

### DIFF
--- a/src/sx-design/.storybook/preview.js
+++ b/src/sx-design/.storybook/preview.js
@@ -4,7 +4,7 @@ import type { Node } from 'react';
 import { MINIMAL_VIEWPORTS } from '@storybook/addon-viewport';
 
 import { SxDesignProvider } from '../index';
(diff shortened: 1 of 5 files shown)


mrtnzlml:
Apart from additional constraints on this design system - it's also a great learning experience. For example, did you know there are [CSS Logical Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) specifically designed for use cases like RTL layouts? So you can/should write the styles a bit differently? SX could help with that a lot. 🤓 

Screenshot 2021-04-21 at 14 56 46

_Source: https://adrianroselli.com/2019/11/css-logical-properties.html_

tbergq (approved):

PR #2271, opened 2021-04-21
Eslint: do not ignore `.storybook` folder

mrtnzlml:
Eslint ignores this folder by default, see: https://github.com/eslint/eslint/issues/8429


src/sx-design/.storybook/manager.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 
+// eslint-disable-next-line import/no-extraneous-dependencies
```
tbergq, line 3: Maybe we should add this file path to the rule in `.eslitrc.js`?

mrtnzlml: Or maybe even to the Eslint Config? I just noticed we already ignore `**/*.stories.js` there: https://github.com/adeira/universe/blob/e506205d1b54f4c63334597b5a584c296747c950/src/eslint-config-adeira/ourRules.js#L607-L618

tbergq: Yeah, I guess that is what I meant, add `**/.storybook/**` to the `devDependency` array.

mrtnzlml: @tbergq I tried multiple combinations of the glob pattern and I am unable to figure out what should it look like to ignore the `.storybook` directory. None of it (even the one suggested by you) works. 🤔

tbergq: It seems to work for me and the CI, I pushed a commit on top of yours 🙂

tbergq: Ok, I think I understand your problem, you were trying in `src/eslint-config-adeira/ourRules.js`, but we also have the rule specified in `.eslintrc.js` which effectively overwrites the `devDependencies` from `ourRules.js`

tbergq: > I pushed a commit on top of yours

I updated the `.eslintrc.js` file, but I was thinking maybe we should rather remove this rule from that file, and only depend on the one in `ourRules.js`, I guess having this rule in both places causes confusion. WDYT?

mrtnzlml: > Ok, I think I understand your problem, you were trying in `src/eslint-config-adeira/ourRules.js`, but we also have the rule specified in `.eslintrc.js` which effectively overwrites the `devDependencies` from `ourRules.js`

Yes! That's it! Thanks for your help, I will check asap.

> I updated the .eslintrc.js file, but I was thinking maybe we should rather remove this rule from that file, and only depend on the one in ourRules.js, I guess having this rule in both places causes confusion. WDYT?

Totally agree. But I will leave it for some (near) future PR. We need to think a bit about breaking changes.

mrtnzlml (commented):
@tbergq Can I ask you for a review, please?

tbergq (approved):

tbergq:
> @tbergq Can I ask you for a review, please?

Looks good to me, if you agree with my changes 🙂

PR #2296, opened 2021-04-26
SX Design: add new `Image` component

mrtnzlml:


src/sx-design/src/Image/Image.js
```diff
@@ -0,0 +1,92 @@
...
+    <div
+      className={styles({
+        background: true,
+        backgroundLight: theme === 'light',
+        backgroundDark: theme !== 'light',
+      })}
+      style={{ width: props.width, height: props.height }}
+    >
+      {isImageLoaded === false && props.blurhash != null ? (
+        <Blurhash hash={props.blurhash} width={props.width} height={props.height} />
```
michalsanger, line 44: This fails badly if the `blurhash` is invalid. I'd recommend validation of error handling

michalsanger (approved):

tbergq:
What about adding support for `webp` images, with fallback?

Doesn't have to be this PR though, just a thought. 

https://css-tricks.com/using-webp-images/

src/sx-design/src/Image/Image.js
mrtnzlml: Good catch! Fixed and added a test case for it. Thanks!

mrtnzlml:
> What about adding support for `webp` images, with fallback?
> 
> Doesn't have to be this PR though, just a thought.
> 
> https://css-tricks.com/using-webp-images/

Good idea. I am happy to add it or accept a PR as long as some SX Design user wants it. YAGNI otherwise. :)

PR #2309, opened 2021-04-30
CI: setup Dockerfile builder

mrtnzlml:
The idea is that `adeira/universe` is responsible for the `Dockerfile` building and pushing so developers do not have to take care of it at all. It uses GitHub Actions build matrix together with `dorny/paths-filter` action to build the Dockerfiles only when needed.

### DIFF
--- a/.github/workflows/dockerfile-builder.yml
+++ b/.github/workflows/dockerfile-builder.yml
@@ -1,0 +1,64 @@
+name: Dockerfile Builder
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  # https://docs.docker.com/ci-cd/github-actions/
+  build-and-publish:
+    strategy:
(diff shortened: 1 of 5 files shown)


.github/workflows/dockerfile-builder.yml
```diff
@@ -0,0 +1,41 @@
+name: Dockerfile Builder
+
+on:
+  push:
+    branches:
+      - master
```
michalsanger, line 6: How about limit it by path of projects with docker file? Or is it cheap to run this workflow on every push to master?

mrtnzlml: I updated the PR, please have a look again. I am filtering the workflow on a job level though so I can combine it easily with the GH build matrices.

michalsanger (approved):

PR #2338, opened 2021-05-04
Shipit: setup exports for kochka.com.mx

mrtnzlml:
It is not a perfect solution to be exporting it to my personal repository, I admit. I will take care of it. I am looking for a better solution but this is the most cost-effective for now.


michalsanger (approved):
Are you sure there is no need for custom config files in exported repo? You know, the stuff we usually have in `__github__` folder.

mrtnzlml:
> Are you sure there is no need for custom config files in exported repo? You know, the stuff we usually have in `__github__` folder.

I actually tried to export it via Shipit already and build it and it worked well. So I think it's not necessary. I might be wrong though in which case I will send follow-up PRs.

Issue #2341, opened 2021-05-05
Road to native ES modules


mrtnzlml:
There are some dependencies that started pushing native ES modules in a very invasive way, see:

- [ ] `strip-ansi` https://github.com/adeira/universe/pull/2249
- [ ] `hast-util-to-html` https://github.com/adeira/universe/pull/2323

We need to replace or upgrade these. However, we are not ready for it. So here is the plan to make it happen:

- [x] cleanup our legacy quirky ES modules system https://github.com/adeira/universe/pull/2339
- [ ] remove `useESModules` from `@babel/runtime` and maybe others, see: https://babeljs.io/blog/2021/02/22/7.13.0#automatic-es-modules-in-babelruntime-12632httpsgithubcombabelbabelpull12632
- [ ] implement conditional packages exports: https://nodejs.org/api/packages.html#packages_conditional_exports and https://nodejs.org/api/packages.html#packages_dual_commonjs_es_module_packages (including support in NPM publisher)
- [ ] read through https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

michalsanger:
So what is our goal? I see two different ones:

1. To be able to use ESM packages within Universe
2. Export libs from Universe to NPM as to support both ESM and CJS

mrtnzlml:
@michalsanger I see both of the points to be goals of this issue. I updated the issue description to be more clear about the goals. Thanks!

PR #2371, opened 2021-05-10
Dependabot: Ignore Refractor updates

michalsanger:
see #2341


mrtnzlml (approved):
I love how the JS community is thinking hard about migration strategies. 🤦

PR #2427, opened 2021-05-24
Abacus BO: use SX Design button in the new forms

mrtnzlml:
I also renamed some base components to be more clear about their purpose.


mrtnzlml:
Hola @adeira/devs! 👋 This needs your input as well because I am editing part of SX Design (`src/sx-design/src/Button/Button.js`). Thanks!

michalsanger (approved):

PR #2435, opened 2021-05-26
Use Jest version 27.0.1

mrtnzlml:
See release notes: https://jestjs.io/blog/2021/05/25/jest-27

There is currently one issue with global `afterEach` callbacks order which I workarounded by using `afterAll` (see: https://github.com/facebook/jest/issues/11456). I will wait for the resolution of that issue and update our code accordingly.


michalsanger (approved):

.jest.config.js
```diff
@@ -27,7 +27,7 @@ const commonProjectConfig = {
   transform: {
     '^.+\\.js$': 'babel-jest',
   },
-  timers: 'fake',
+  timers: 'legacy', // TODO: migrate to the Jest default values!
```
mrtnzlml, line 30: https://github.com/adeira/universe/issues/2436

PR #2504, opened 2021-06-09
Shipit: fix patch rendering with "---" in description

mrtnzlml:
I noticed that Dependabot commits are not being exported correctly (see https://github.com/adeira/universe/commit/cc7b3818e06f95c2732f75f165a2b98c6eeab135 vs. https://github.com/adeira/eslint-config-adeira/commit/0948aa9822a603d1001d4911865ac3dfb9bed309). It started happening just a few days ago after Dependabot changed their commit message format.

Turns out that "---" in a description has a special meaning when applying the changes via `git am` (that's what Shipit does internally). Specifically, it removes this part of the messages which removes `adeira-source-id` as well making future Shipit exports broken (Shipit cannot find the `adeira-source-id` handle so it tries to apply the same changes again resulting in a patch conflict).

This change workarounds the issue by prefixing "---" with one space (the same fix as in `facebook/fbshipit`). It's not perfect but it does the job.

I also created a new integration test which takes any existing commit in `adeira/universe` and tries to parse it, filter it and render it (while snapshoting each of the phases). I created it in a hope of finding the issue but that didn't happen (it would require to actually apply the patch which is impossible when there is no previous project history and that's complicated to setup).

Followup steps: we need to marge this fix and then I will reset the incorrectly exported commits in every affected repository. Shipit should then start exporting the changes correctly again.

### DIFF
--- a/src/monorepo-shipit/src/__tests__/__snapshots__/Integration.test.js.snap
+++ b/src/monorepo-shipit/src/__tests__/__snapshots__/Integration.test.js.snap
@@ -1,0 +1,273 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`creates correct changeset from commit cc7b3818e06f95c2732f75f165a2b98c6eeab135, filters it and renders it as expected: 1 - parsed changeset without filters 1`] = `
+Changeset {
+  "author": "\\"dependabot[bot]\\" <49699333+dependabot[bot]@users.noreply.github.com>",
+  "coAuthorLines": Array [],
+  "debugMessages": Array [],
+  "description": "Bumps [prettier](https://github.com/prettier/prettier) from 2.3.0 to 2.3.1.
+- [Release notes](https://github.com/prettier/prettier/releases)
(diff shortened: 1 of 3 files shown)


src/monorepo-shipit/src/__tests__/__snapshots__/Integration.test.js.snap
```diff
@@ -0,0 +1,273 @@
...
+From: \\"dependabot[bot]\\" <49699333+dependabot[bot]@users.noreply.github.com>
+Date: Mon, 7 Jun 2021 05:12:25 +0000
+Subject: [PATCH] Bump prettier from 2.3.0 to 2.3.1
+
+Bumps [prettier](https://github.com/prettier/prettier) from 2.3.0 to 2.3.1.
+- [Release notes](https://github.com/prettier/prettier/releases)
+- [Changelog](https://github.com/prettier/prettier/blob/main/CHANGELOG.md)
+- [Commits](https://github.com/prettier/prettier/compare/2.3.0...2.3.1)
+
+ ---
```
mrtnzlml, line 156: This is how the workaround will be exported.

michalsanger (approved):

PR #2524, opened 2021-06-10
CI: Rust CI path filter

michalsanger:
Run CI for Rust projects only when there are changed files in the pull request

### DIFF
--- a/.github/workflows/continuous-integration-rust.yml
+++ b/.github/workflows/continuous-integration-rust.yml
@@ -24,8 +24,17 @@ jobs:
     steps:
       - uses: actions/checkout@v2.3.4
 
+      - uses: dorny/paths-filter@v2
+        id: changes
+        with:
+          filters: |
+            src:
+              - 'src/signedsource-rs/**'
+              - 'src/abacus/**'
+
       # https://github.com/actions-rs/toolchain
       - uses: actions-rs/toolchain@v1.0.7
+        if: steps.changes.outputs.src == 'true'
         with:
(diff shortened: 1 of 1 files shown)


michalsanger:
Not sure this will work because skipped checks are marked as required...

mrtnzlml:
> Not sure this will work because skipped checks are marked as required...

@michalsanger I am afraid it won't work. However, you can use [dorny/paths-filter](https://github.com/dorny/paths-filter) to workaround this issue which should work fine. We use it in Dockerfile builder: https://github.com/adeira/universe/blob/00413dbb8c38fe48e7476a48d51486121adb8e5b/.github/workflows/dockerfile-builder.yml#L37-L42

michalsanger:
`dorny/paths-filter` looks good. Just to be sure, could you try to approve this PR?

mrtnzlml (approved):

michalsanger:
WTF kodiak! :-)

.github/workflows/continuous-integration-rust.yml
```diff
@@ -24,8 +24,17 @@ jobs:
     steps:
       - uses: actions/checkout@v2.3.4
 
+      - uses: dorny/paths-filter@v2
+        id: changes
+        with:
+          filters: |
+            src:
+              - 'src/signedsource-rs/**'
+              - 'src/abacus/**'
```
mrtnzlml, line 33: One slight improvement for the future would be to use the test matrix above.

PR #2527, opened 2021-06-10
Universe: rename `kochka.com.mx` to `abacus-kochka`

mrtnzlml:
First, there are some problems with tests and I suspect that the project name might be the culprit. Moreover, `.com.mx` is naming might not reflect the future plans and it's better not to encode the domain name in the folder structure.


mrtnzlml:
@adeira/devs Can I please ask you for a quick approval? I am simply renaming a directory under `src/`. Thanks!

tbergq (approved):

PR #2619, opened 2021-06-28
SX Design: use rounded edges (`Image`, `ProductCard`)

mrtnzlml:
Using rounded edges makes it more uniform with the rest of the design system. I also removed `squared` option from `Skeleton` component. It's because this option is not being used and SX Design should implement only what's actually needed and used.

### DIFF
--- a/src/sx-design/src/Image/Image.js
+++ b/src/sx-design/src/Image/Image.js
@@ -6,8 +6,6 @@ import sx from '@adeira/sx';
 import { Blurhash } from 'react-blurhash';
 import { isBlurhashValid } from 'blurhash';
 
-import useSxDesignContext from '../useSxDesignContext';
-
 type Props = {
   +'width': number,
   +'height': number,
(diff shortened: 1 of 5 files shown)


src/sx-design/src/Image/Image.js
```diff
@@ -76,18 +72,15 @@ const styles = sx.create({
   background: {
     display: 'flex',
     flexDirection: 'column',
-  },
-  backgroundLight: {
-    backgroundColor: 'lightgrey',
-  },
-  backgroundDark: {
-    backgroundColor: 'grey',
+    backgroundColor: 'rgba(var(--sx-accent-2))',
```
mrtnzlml, line 75: This is why it's important to prefer colors from SX Design instead of re-inventing the wheel: `--sx-accent-2` is already light/dark theme sensitive and there is no need for manual custom colors based on the theme. It makes the component implementation much nicer.

michalsanger (approved):

PR #2625, opened 2021-06-28
SX Design: automatically set `font-family`

mrtnzlml:


src/sx-design/src/SxDesignProvider.js
```diff
@@ -98,7 +98,10 @@ export default function SxDesignProvider(props: Props): Node {
 
 /* eslint-disable sx/valid-usage */
 const styles = sx.create({
-  common: SxDesignProviderCSSVariables.common,
+  common: {
+    ...SxDesignProviderCSSVariables.common,
+    fontFamily: 'sans-serif',
```
michalsanger, line 103: How about `ui-sans-serif`?

mrtnzlml: I've changed it to `ui-sans-serif, sans-serif` because only `ui-sans-serif` is widely unsupported: https://caniuse.com/extended-system-fonts

michalsanger: In that case, maybe `ui-sans-serif, system-ui, sans-serif`?

michalsanger: https://caniuse.com/?search=system-ui

mrtnzlml: Your comments motivated me to dig a bit deeper into this problem and I choose to copy `font-family` used here on GitHub.com. It looks the best from what I tried. Hopefully, this is the best what we can do now. Please, have a look again.

michalsanger (approved):

PR #2627, opened 2021-06-28
SX Design: implement `findBestTooltipPosition`

mrtnzlml:
This allows us to position to `Tooltip` component better when there is not enough space.

### DIFF
--- a/src/sx-design/src/SxDesignPortal.js
+++ b/src/sx-design/src/SxDesignPortal.js
@@ -8,20 +8,23 @@ type Props = {
   +children: Node,
 };
 
-export default function TooltipPortal(props: Props): Portal | null {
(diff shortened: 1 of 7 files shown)


src/sx-design/src/SxDesignPortal.js
```diff
@@ -8,20 +8,23 @@ type Props = {
   +children: Node,
 };
 
-export default function TooltipPortal(props: Props): Portal | null {
+const REACT_PORTAL_SELECTOR = '#sx-design-react-portal-root';
```
michalsanger, line 11: Maybe export it and use it in `src/sx-design/src/SxDesignProvider.js`?

michalsanger (approved):

src/sx-design/src/SxDesignPortal.js
mrtnzlml: Good idea! 👍

PR #2643, opened 2021-06-30
SX Design: add new component `MissingData`

mrtnzlml:
This is especially useful when you want to express that some part of the data is missing or cannot be loaded.

TODO:

- [x] regenerate FBT


mrtnzlml:
I forgot to mention that I also modified the `Tooltip` component interface to be more similar to `<abbr />`.

michalsanger (approved):

PR #2664, opened 2021-07-01
CI: enable Yarn dependencies cache

mrtnzlml:
See: https://github.com/actions/setup-node/releases/tag/v2.2.0


michalsanger (approved):
Let's try. I already tried in different repo, it created the cache, hit it, but no improvement on speed because the cache size was 0 MB :-(

PR #2668, opened 2021-07-01
SX: expand logical CSS borders

mrtnzlml:


mrtnzlml:
_I think_ this actually doesn't do anything in reality because `expandShorthandProperties` won't let the logical border through. We need to refactor the code/tests a little it. 🤔

michalsanger (approved):

mrtnzlml:
@michalsanger Would you please check it again? I improved the types and added some extra tests to make sure these functions are actually applied as expected.

michalsanger (approved):

PR #2669, opened 2021-07-01
SX Eslint: suggest logical margins and paddings

mrtnzlml:


src/abacus-kochka/src/Layout.js
```diff
@@ -97,7 +94,6 @@ const styles = sx.create({
   },
   footer: {
     color: 'rgba(var(--font-color-dark))',
-    paddingLeft: '1rem',
-    paddingRight: '1rem',
+    paddingInline: '1rem',
```
mrtnzlml, line 97: This is another advante of logical CSS properties - it's shorter (`paddingInline` instead of `paddingLeft` + `paddingRight`).

michalsanger (approved):

PR #2690, opened 2021-07-05
CI: Enable yarn cache

michalsanger:
Should work now:
https://github.com/actions/setup-node/issues/281#issuecomment-873961884


.github/workflows/continuous-integration-javascript.yml
```diff
@@ -4,7 +4,6 @@ on: [push]
 
 env:
   CI: true
-  FORCE_COLOR: '1' # https://github.com/chalk/chalk/issues/203
```
mrtnzlml, line 7: I think we should clean up `src/sx-tailwind-website/__github__/.github/workflows/continuous-integration.yml` as well.

mrtnzlml (approved):

PR #2695, opened 2021-07-06
SXT Web: Fix CI in exported project

michalsanger:
Cache can't be used in exported project becuse `yarn.lock` is missing:
https://github.com/adeira/sx-tailwind-website/runs/2996597046#step:3:15


mrtnzlml (approved):
Makes sense.

PR #2704, opened 2021-07-07
Abacus: improve `resolve_aql` interface

mrtnzlml:
I believe this is easier to use.


mrtnzlml:
I will double-check the changes tomorrow when I am less tired.

mrtnzlml:
OK, Rust macros are awesome! ❤️

PR #2724, opened 2021-07-08
SX Design: make `Image` and `ProductCard` responsive

mrtnzlml:
This change is needed to accommodate smaller screens where the size of the product card should be smaller. Both `Image` and `ProductCard` components are now rendering into the parent component with aspect ratio 1:1. This way the users of this library can resize these components however they want. Aspect ratio 1:1 is currently hardcoded but can, of course, be changed when there is a need for it.


src/sx-design/src/ProductCard/ProductCard.js
```diff
@@ -45,51 +45,53 @@ type Props = {
...
 
-  const DEFAULT_HEIGHT = 250;
-  const DEFAULT_WIDTH = 250;
-
   return (
     <div
-      style={{ height: DEFAULT_HEIGHT, width: DEFAULT_WIDTH }}
-      className={styles('wrapper')}
+      className={styles('wrapper', 'aspectRatioBox')}
       onMouseOver={() => setIsHovered(true)}
```
michalsanger, line 51: not related, but maybe
```suggestion
      onMouseEnter={() => setIsHovered(true)}
```

michalsanger (approved):

PR #2726, opened 2021-07-11
SX: update SSR not to depend on Next.js

mrtnzlml:
This change replaces `renderPageWithSX` function that is Next.js dependent with `getStyleTag` which should work anywhere (it only returns `<style />` tag expected by SX).


michalsanger (approved):
How about bump the version?

mrtnzlml:
> How about bump the version?

I added an extra commit releasing 0.27.0. In case you are trying this project somewhere, please note that it's still unstable and subject to change. I do not expect anything dramatic but there are still several changes in the queue. 😎

PR #2741, opened 2021-07-15
Colors: add new function `chooseHigherContrast`

mrtnzlml:


michalsanger (approved):
Would be great to support alpha channel:
http://marcodiiga.github.io/rgba-to-rgb-conversion

mrtnzlml:
> Would be great to support alpha channel:
> http://marcodiiga.github.io/rgba-to-rgb-conversion

This is cool. The missing alpha channel is something I'd like to change for sure. It is a design mistake of this package that it uses RGB triplet `[r, g, b]` and it ignores alpha values. 🙈 Let's stick with it for now, but I am already thinking about how to refactor it.

PR #2749, opened 2021-07-20
Colors: improve RGB(A) and HSL(A) detection

mrtnzlml:
The goal of this change is to follow specification and how browsers are actually understanding colors. Basically, all previously valid colors are still valid, but the check is no longer benevolent. I verified all invalid cases in the browser to make sure they also do not accept these invalid colors.


mrtnzlml:
Alternatively, I also tried (and I am still trying) https://github.com/d3/d3-color, however, it seems like it's having many false positives (it rejects many valid color formats). It is a great inspiration nonetheless and I'd like to achieve a similar interface.

tbergq (approved):

PR #2154, opened 2021-04-06
Upgrade to Yarn Berry

mrtnzlml:
🚧 **WIP - this PR is here just to see what it takes to migrate to Yarn Berry** 🚧

See: https://yarnpkg.com/getting-started/migration

Closes: https://github.com/adeira/universe/issues/1734

PR #2794, opened 2021-07-26
Universe: remove Bazel files

mrtnzlml:
The reasoning is fairly simple: nobody is actively working on
integrating with Bazel and there are no plans/capacity to do so. It
might be because Bazel is quite complex, and it's hard to set up it in a
way we need it to be. Moreover, it's blocking other efforts like this
one for example: https://github.com/adeira/universe/pull/2154 (Yarn 2).

Note: this doesn't mean that there is no need for something like Bazel.
However, it doesn't have to be Bazel (preferably something simpler that
does the job well enough).


mrtnzlml:
This is actually the first commit in a series of cleanup commits. Specifically, Shipit needs some cleaning, however, I think it should happen after this one.

tbergq (approved):

PR #2802, opened 2021-07-26
SX Design: add missing `Text` tests

mrtnzlml:


mrtnzlml:
🤔 I think I messed up this PR when I was separating the changes locally. We should probably first merge https://github.com/adeira/universe/pull/2801 which will fix the failing tests.

tbergq (approved):

PR #2822, opened 2021-07-27
SX Design: add new props `as`, `size`, and `weight` to `Text` component

mrtnzlml:
Additionally, I added a deprecation note to `Heading` and `Section`
components because I think they bring unnecessary complexity to the
system, and it's better to replace them with the improved `Text`
component.


mrtnzlml:
I forgot to mention that it's inspired by https://vercel.com/design/text as usual.

tbergq (approved):

PR #2826, opened 2021-07-27
SX Design: allow `Text` component to be rendered as `small`

mrtnzlml:


src/sx-design/src/Text/Text.js
```diff
@@ -76,7 +78,7 @@ export default function Text(props: Props): Node {
         w950: props.weight === 950,
       })}
       style={{
-        color: props.backgroundRef != null ? accessibleTextColor : 'rgba(var(--sx-foreground))',
```
mrtnzlml, line 79: Text should inherit the color of the surrounding markup color (unless overwritten). This was a mistake.

tbergq (approved):

Issue #1854, opened 2021-02-23
Babel presets 'next/babel' and '@adeira/babel-preset-adeira' cannot be used together


mrtnzlml:
This is a follow-up after these PRs/issues:

- https://github.com/adeira/universe/commit/1c5d0b46ea5f0814a6d3b3f1566c2b7d8d68e908
- https://github.com/adeira/universe/pull/1805

Basically, we have issues using these two presets together (for example, crashing on `Missing class properties transform` when running `yarn workspaces @adeira/example-relay test-bc` and with the explicit `@babel/plugin-proposal-class-properties` we cannot write `declare` in our JS code).

This possibly requires creating a small reproducible code and getting in touch with Next.js team to figure out what do to.

PR #2867, opened 2021-08-02
Relay example: update Babel configs

mrtnzlml:
With a better explanation and for OSS as well. Hopefully, this will
unblock the Vercel builds again.

Related issue: https://github.com/adeira/universe/issues/1854


src/example-relay/.babelrc.js
```diff
@@ -1,13 +1,11 @@
...
-  /**
-   * test-bc script broke after upgrading some babel packages.
-   * Adding @babel/plugin-proposal-class-properties made it work, but it should not be necessary
-   * to add it here, since we already have it in @adeira/babel-preset-adeira.
-   * We need to leave it here for now for the test-bc script to work, but we should try
-   * to remove it in the future
-   */
-  plugins: ['relay', '@babel/plugin-proposal-class-properties'],
+  presets: ['@adeira/babel-preset-adeira'],
+  // For some unknown reason, we have to specify here the `class-properties` Babel plugin explicitly
```
tbergq, line 5: Can we remove this comment now?

tbergq (approved):

src/example-relay/.babelrc.js
mrtnzlml: I don't think so. Previously there was a comment about the `test-bc` script that doesn't exist anymore. But the issue is still valid and it affects the application itself as well. And it's still unclear to me why these two presets interfere, see: https://github.com/adeira/universe/issues/1854.

PR #2870, opened 2021-08-03
Relay example: use SX Design

mrtnzlml:
There are many other parts of the application that could be written
better and with more of the SX Design. However, this should be a good
start.


src/example-relay/src/Homepage/locations/LocationsPaginated.js
```diff
@@ -52,7 +52,8 @@ export default function LocationsPaginated(props: Props): Node {
           <Location key={edge?.node?.id} location={edge?.node} />
         ))}
       </LocationList>
-      <Button onClick={loadMore} disabled={!hasNext}>
+      <Button onClick={loadMore} isDisabled={!hasNext}>
+        {/* $FlowFixMe[incompatible-type]: Button children should be FBT */}
```
mrtnzlml, line 56: We need to think about this type in SX Design. The goal is to encourage well-built (and therefore translated) applications so maybe we can use here FBT as well. 🤔

michalsanger: That tight coupling with FBT?

michalsanger (approved):

src/example-relay/src/Homepage/locations/LocationsPaginated.js
mrtnzlml: I will relax the `Button` types in a follow-up PR.

mrtnzlml: https://github.com/adeira/universe/pull/2877

PR #2886, opened 2021-08-04
Eslint config: refactor internal structure

mrtnzlml:
The purpose of this change is to modify internal structure of the
config. Specifically, it moves some files under `src/`, separates
presets into individual files, dedupes necessary plugins for each preset
and fixes some minor issues which were uncovered during this
refactoring.

There are still some parts that should be fixed (for example do not
include prettier in every preset), however, I wanted to keep the
snapshot changes on minimum so it's clear what changed.

I will follow with additional cleanup commits later.


src/eslint-config-adeira/__tests__/__snapshots__/presets.test.js.snap
```diff
@@ -571,6 +571,7 @@ Object {
   "plugins": Array [
     "eslint-plugin-flowtype",
     "eslint-plugin-fb-flow",
+    "eslint-plugin-prettier",
```
mrtnzlml, line 574: This was somehow strangely broken in the previous presets: the prettier rules are being used but the plugin was not specified anywhere.

michalsanger (approved):

PR #2562, opened 2021-06-17
Eslint config: add new `@next/eslint-plugin-next`

mrtnzlml:
TODO: make it configurable (should not be applied always but only for nextjs projects)

See:

- https://github.com/adeira/universe/discussions/2560
- https://github.com/vercel/next.js/discussions/11731
- https://nextjs.org/docs/basic-features/eslint

PR #2897, opened 2021-08-05
Eslint config: add new Next.js subset

mrtnzlml:
Supersedes https://github.com/adeira/universe/pull/2562


mrtnzlml:
Is this how you envisioned it @michalsanger?

michalsanger (approved):
My idea was to use `@next/eslint-plugin-next` directly but this is even better.

PR #2914, opened 2021-08-09
SX Design: change Storybook links to docs in README

mrtnzlml:
It's better to point directly to the component docs because it contains
additional useful information that would be otherwise easy to miss.

### DIFF
--- a/src/sx-design/README.md
+++ b/src/sx-design/README.md
@@ -74,35 +74,35 @@ Legend:
 | [`<Text />`]          |     ✅      |     ✅      |      ✅       |    ✅    |  🧐   |
 | [`<Tooltip />`]       |     ✅      |     ✅      |      ✅       |    ✅    |  🧐   |
 
-[`<badge />`]: https://sx-design.vercel.app/?path=/story/components-badge
-[`<button />`]: https://sx-design.vercel.app/?path=/story/components-button
-[`<buttonlink />`]: https://sx-design.vercel.app/?path=/story/components-buttonlink
-[`<emoji />`]: https://sx-design.vercel.app/?path=/story/components-emoji
-[`<entity />`]: https://sx-design.vercel.app/?path=/story/components-entity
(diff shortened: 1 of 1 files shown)


michalsanger (commented):
> After this change, docs tab will always be opened by default

I can't say this works for me. I run `yarn storybook` in `src/sx-design`, the Storybook opens automatically but the active tab is "Canvas".

mrtnzlml:
> > After this change, docs tab will always be opened by default
> 
> I can't say this works for me. I run `yarn storybook` in `src/sx-design`, the Storybook opens automatically but the active tab is "Canvas".

@michalsanger Is it possible that it doesn't work when you open it for the first time? Does it work when you navigate around the stories? I've seen some issues around the first opening - maybe "always" is more like "almost always"?

mrtnzlml:
For example: https://github.com/storybookjs/storybook/issues/13128

michalsanger:
When I select the Docs tab it keeps selected across other stories, but when I open the URL again, it's back on Canvas.

mrtnzlml:
Can you please make sure that you are not opening an URL having `/story/`? This is how it works for me:

- open [localhost:6006](http://localhost:6006) (opens `http://localhost:6006/?path=/story/styleguide-colors--page` which is wrong but whatever)
- click on `Text` tab (opens `http://localhost:6006/?path=/docs/components-text--default` ✅)

🤔

michalsanger:
When I open http://localhost:6006 it redirects to http://localhost:6006/?path=/story/styleguide-colors--page and the Canvas tab is active. When I click on Text story in left menu, I get http://localhost:6006/?path=/story/components-text--default and the Canvas tab is active :-(

mrtnzlml:
OK, I will check later what could be the problem. Thanks.

mrtnzlml:
@michalsanger Can you please have a look again? I didn't figure out why was it broken so I changed the PR to simply point to the docs directly from README. That should be enough.

michalsanger (approved):

PR #2918, opened 2021-08-09
Relay example: improve main navigation styles

mrtnzlml:
The main goal of this change was to use `Text` component for the
navigation so we don't have to deal with the links color. This required
some small changes in `Text` component types as well.

Additionally, I slightly update `Link` styles so the navigation (or any
links, really) looks even better.


src/sx-design/src/Text/Text.js
```diff
@@ -1,36 +1,59 @@
 // @flow
 
-import React, { type Node } from 'react';
+import React, { type Node, type Element } from 'react';
 import sx from '@adeira/sx';
 
 import useAccessibleColors from './useAccessibleColors';
 
+// https://developer.mozilla.org/en-US/docs/Web/CSS/font-size
+export type TextSupportedSize = 10 | 12 | 14 | 16 | 20 | 24 | 32 | 40 | 48;
```
michalsanger, line 9: What's the unit here?

michalsanger (approved):

PR #2919, opened 2021-08-09
SX: expand logical shorthand margins and paddings

mrtnzlml:


src/sx-design/src/test-utils.js
```diff
@@ -39,9 +39,9 @@ const renderWithProviders = (
...
 };
 
-// re-export everything
+// re-export needed things without modifications
 // eslint-disable-next-line import/no-extraneous-dependencies
-export * from '@testing-library/react';
+export { fireEvent } from '@testing-library/react';
 
-// override render method
+// export custom render method(s)
```
mrtnzlml, line 46: Seems like we merged some Eslint plugin upgrade and now it complains about overwriting the exports. This is better.

michalsanger (approved):

PR #2942, opened 2021-08-12
SX Design: improve `ProductCard` styles

mrtnzlml:
First, seems like I broke the styles a little when I was removing the
deprecated `Heading` component. This change fixes it.

Second, the card looked a bit weird when the product names are long.
So I tried to improve it a little by introducing variable title
size and limiting the width of the highlight.


mrtnzlml:
This is how it was broken before:

Screen Shot 2021-08-11 at 20 44 35

This is how it looked like before for long titles:

Screen Shot 2021-08-11 at 20 44 44

And after this change (notice the variable title sizes):

Screen Shot 2021-08-11 at 21 30 11

Better. 🙂

michalsanger (approved):

PR #3000, opened 2021-08-27
SX Design: enforce `Button` ARIA label for non-textual children

mrtnzlml:
Normally, ARIA label would be derived from the `Button` children text.
However, this is not the case when the children is just an `Icon` since
there is no text to start with.

In this case we enforce to specify the ARIA label manually.


src/sx-design/src/Button/Button.js
```diff
@@ -18,12 +19,28 @@ type Props = {
...
+  +'aria-label'?: Fbt,
 };
 
 /**
  * Creates a traditional HTML button with the correct attributes.
  */
 export default function Button(props: Props): React.Element<'button'> {
+  const childrenCount = React.Children.count(props.children);
+  if (childrenCount === 1) {
+    React.Children.forEach(props.children, (child) => {
```
michalsanger, line 31: Feels strange to use `forEach` when you know there is only one child. Wouldn’t it be more readable with `React.Children.only`?

mrtnzlml: I agree this is a bit awkward. Unfortunately, `React.Children.only` tests whether there is children is a React Element and it doesn't accept any React Node (unlike `React.Children.forEach`). In other words:

```js
React.Children.only(<div>test</div>); // OK
React.Children.only('test'); // Error: React.Children.only expected to receive a single React element child.

React.Children.forEach(<div>test</div>, () => {}); // OK
React.Children.forEach('test', () => {}); // OK
```

We have to be able to accept just a string though (`<Button>test</Button>`). So this would not work:

```js
const childrenCount = React.Children.count(props.children);
if (childrenCount === 1) {
  const child = React.Children.only(props.children);
  if (child.type === Icon) {
    // This is a case where the only Button children is an Icon. Normally, the ARIA label
    // would be derived from the Button text, but this special case breaks accessibility
    // since there is no text to start with.
    invariant(
      props['aria-label'] != null,
      'Icon is the only Button child so ARIA label property must be specified.',
    );
  }
}
```

Am I missing something?

michalsanger (approved):

PR #3033, opened 2021-08-31
SX Design: do not publish storybook related files

mrtnzlml:


src/sx-design/src/Modal/Modal.js
```diff
@@ -33,7 +33,7 @@ type Props = {
  * element that was focused just before the modal activated.
  *
  * TODO: implement important accessibility features:
- *  - correct aria attributes
+ *  - prevent background scroll
```
mrtnzlml, line 36: This is just so I don't forget. The ARIA attributes should be set properly already.

mrtnzlml: Fuck it, better do it properly: https://github.com/adeira/universe/pull/3034

michalsanger (approved):

PR #3044, opened 2021-09-02
Relay: use version 12.0.0

mrtnzlml:
See: https://github.com/facebook/relay/releases/tag/v12.0.0

---

A big part of Relay 12 is focused on logging. So does this PR: it has 3 commits where the first one is removing the old loggers, the second one upgrades Relay itself and the last commit implements the logger again (but hopefully better than before).

I would like to ship additional breaking changes before releasing a new major `@adeira/relay` version, stay tuned.


src/relay/src/requestSubscription.js
```diff
@@ -1,17 +0,0 @@
-// @flow
-
-import { requestSubscription as _requestSubscription, type Environment } from 'react-relay';
-import type { GraphQLSubscriptionConfig, Disposable } from 'relay-runtime';
-
-type SubscriptionPayload = { +[key: string]: any, ... };
```
mrtnzlml, line 6: Something got fixed in Relay so this is no longer needed.

src/relay/src/RelayLogger.js
```diff
@@ -0,0 +1,78 @@
...
+      const seconds = Math.round((Date.now() - fetchTime) / 1000);
+      return `🟢 ${seconds}s old${shouldFetch ? ', re-fetching' : ''}`;
+    }
+    return `🟢 ${shouldFetch ? 're-fetching' : ''}`;
+  } else if (queryAvailability.status === 'stale') {
+    return `🟠 ${shouldFetch ? 're-fetching' : ''}`;
+  } else if (queryAvailability.status === 'missing') {
+    return `🔴 ${shouldFetch ? 'fetching' : ''}`;
+  }
+  (queryAvailability.status: empty);
```
michalsanger, line 24: All clear, I just don't get this line

michalsanger (approved):

src/relay/src/RelayLogger.js
mrtnzlml: It is an exhaustive check: https://mrtnzlml.com/docs/flow/patterns/exhaustive-checking

Basically, in case I'd forget to implement one of the statuses (or a new one will be added in the future) it will throw the following error:

Screen Shot 2021-09-06 at 8 46 17

In other words, it says: the status was not completely exhausted and `stale` needs to be implemented.

PR #3074, opened 2021-09-05
SX Design: finish `Tabs` accessibility implementation

mrtnzlml:
This commit corrects accessibility implementation of the `Tabs`
component. Basically, you can focus the active tab via `Tab` key and
then you can navigate the tabs using arrows (not via `Tab` key).

This implementation should follow recommendations as per: https://www.w3.org/TR/wai-aria-practices/#tabpanel

### DIFF
--- a/src/sx-design/src/Tabs/Tabs.js
+++ b/src/sx-design/src/Tabs/Tabs.js
@@ -1,9 +1,10 @@
 // @flow
 
-import React, { type Node } from 'react';
+import React, { useRef, type Node } from 'react';
 import sx from '@adeira/sx';
 
 import Text from '../Text/Text';
+import useKeyPress from '../useKeyPress';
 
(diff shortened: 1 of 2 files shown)


src/sx-design/src/Tabs/Tabs.js
```diff
@@ -17,30 +17,96 @@ type Props = {
...
+          // focus FIST tab (going from the right card to the beginning)
+          tabRefs.current[0]?.focus();
+        }
+      }
+    };
+
+    window.addEventListener('keydown', keypressHandler);
+    return () => {
+      window.removeEventListener('keydown', keypressHandler);
+    };
```
mrtnzlml, line 86: TODO: it would be nice to reuse `useKeyPress.js` from https://github.com/adeira/universe/pull/3076

michalsanger (approved):

PR #3081, opened 2021-09-07
eslint-config: set jsx-a11y/accessible-emoji to off

tbergq:
From some testing on os x with VoiceOver, it seems the guidance
provided in the rule jsx-a11y/accessible-emoji leads to a degraded
experience with the label not being announced to the user.

See [issue here](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/627)


michalsanger (approved):

mrtnzlml:
Nice! Thank you.

PR #3126, opened 2021-09-15
Universe: remove `scripts/generateFlow.js`

mrtnzlml:
This script made sense before where we were managing multiple
`.flowconfig` files, but it doesn't bring anything to the table now.
I do have some future plans which would make this script relevant again
but no capacity for it. So, at this moment, it's just an annoyance.


.flowconfig
```diff
@@ -1,49 +1,70 @@
-# @generated SignedSource<<2939b4229c2154e42c786ee5e0eb9e53>>
-#
-# To regenerate run:
-# yarn monorepo-babel-node scripts/generateFlow.js
-#
-[declarations]
-.+/node_modules/.+
-
+# https://flow.org/en/docs/config/ignore/
```
mrtnzlml, line 1: This is essentially a copy of `.flowconfig.template.base`.

PR #3183, opened 2021-09-27
SX Design: add new component `Menu`

mrtnzlml:
Please note: this component is not finished yet, but it's a good start.
I need to focus on correct positioning and accessibility in the future
commits.

Inspiration:

- https://seek-oss.github.io/braid-design-system/components/MenuItem
- https://vercel.com/design/menu

---

Screen Shot 2021-09-27 at 10 10 34


src/sx-design/src/Menu/Menu.js
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * TODO: render in a `Drawer` for small viewports (see `ModalDrawer`)
+ * TODO: use positioning algorithm similar to `Tooltip` (reuse)
+ * TODO: accessibility (https://www.w3.org/TR/wai-aria-practices-1.1/#menu)
+ */
+export default function Menu(props: Props): Node {
+  const [isMenuOpen, setIsMenuOpen] = useState(false);
+
+  const handleButtonClick = useCallback(() => {
```
tbergq, line 30: Why are you using `useCallback` here?

src/sx-design/src/Menu/__tests__/Menu.test.js
```diff
@@ -0,0 +1,61 @@
...
+      </Menu.Item>
+      <Menu.ItemDivider data-testid="MenuItemDivider" />
+      <Menu.Item tint="error" onClick={jest.fn()}>
+        Delete
+      </Menu.Item>
+    </Menu>,
+  );
+
+  // Everything is hidden by default (not in the DOM):
+  expect(queryByText('One')).toBeNull();
```
tbergq, line 27: What about using [toBeInTheDocument](https://testing-library.com/docs/guide-disappearance/#nottobeinthedocument)

src/sx-design/src/Menu/__tests__/Menu.test.js
```diff
@@ -0,0 +1,61 @@
...
+  expect(queryByText('Two')).toBeNull();
+  expect(queryByText('Three')).toBeNull();
+  expect(queryByTestId('MenuItemDivider')).toBeNull();
+  expect(queryByText('Delete')).toBeNull();
+
+  // We have to first open the menu:
+  fireEvent.click(getByRole('button'));
```
tbergq, line 34: Probably not the most important, but there is a package that is better for user events https://testing-library.com/docs/ecosystem-user-event/

src/sx-design/src/Menu/Menu.js
mrtnzlml: "Just in case we are going to optimize `Button` one day" kind of thinking.

But you are right it's probably not a good idea to do it for no reason. Will remove (as well as from the future PRs where I did something simialr).

tbergq (approved):

src/sx-design/src/Menu/__tests__/Menu.test.js
mrtnzlml: I will have a look at it thanks! Since I am not familiar with this package: do you know how does it differ from the `fireEvent`? In other words: why does `fireEvent` even exist if it's better to use something else? 🤔

src/sx-design/src/Menu/__tests__/Menu.test.js
mrtnzlml: I was not aware of this function, thanks!

This brings a question though. I was always using something like this in my tests:

```js
expect(getByText('Click me!')).toBeDefined();
```

Are you saying it's better to use this?

```js
expect(getByText('Click me!')).toBeInTheDocument();
```

And of course, if yes, do you know why? What is the difference? 🤔 

Thanks!

src/sx-design/src/Menu/__tests__/Menu.test.js
tbergq: I think userEvent also triggers hover, mouseOver etc before clicking the button, while the fireEvent does not. I think they were planning on merging this into the main package at some point. 

I'll see if I can find the article for you once I have time.

mrtnzlml: I am already preparing a PR to change the whole package from `fireEvent` to `userEvent`! 👍

src/sx-design/src/Menu/__tests__/Menu.test.js
tbergq: I didn't dive into the specifics, but at least from my pov it reads better. 

There might be some other benefit/edge case it covers 🤷‍♂️

PR #3191, opened 2021-09-28
SX Design: improve `Button` accessibility

mrtnzlml:
Basically, do not use `disabled` attribute for buttons because it makes
worse experience for people with visual impairment, see: https://css-tricks.com/making-disabled-buttons-more-inclusive/

Additionally, I fixed a bug (somehow related) in `LinkButton` so that
disabled link button doesn't trigger `onClick` callback and doesn't
try to navigate to the `href`.


src/sx-design/src/Button/Button.js
```diff
@@ -43,13 +49,21 @@ export default function Button(props: Props): React.Element<'button'> {
     });
   }
 
+  const handleOnClick = useCallback(
```
tbergq, line 52: Why do we need this `useCallback`?

src/sx-design/src/Button/Button.js
```diff
@@ -43,13 +49,21 @@ export default function Button(props: Props): React.Element<'button'> {
...
+    (event) => {
+      if (props.isDisabled !== true) {
+        props.onClick(event);
+      }
+    },
+    [props],
```
tbergq, line 58: If we do need it, the dependency array should probably only include the two attributes being used in the function.

src/sx-design/src/Link/LinkButton.js
```diff
@@ -27,8 +27,25 @@ type Props = {
...
+export default (React.forwardRef(function LinkButton(props, ref): Node {
+  const handleOnClick = useCallback(
+    (event) => {
+      if (props.isDisabled !== true) {
+        props.onClick?.(event);
+      } else {
+        event.preventDefault();
+      }
+    },
+    [props],
```
tbergq, line 46: Same questions as above

tbergq:
Thanks for sharing that article, TIL 🙂

src/sx-design/src/Button/Button.js
mrtnzlml: Like with the `Menu` PR (and https://github.com/adeira/universe/pull/3192) - the `useCallback` is incorrect here, thanks for pointing it out.

mrtnzlml:
> Thanks for sharing that article, TIL 🙂

BTW please note that it's not so simple: https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_disabled_controls

Basically, there are cases where it might have sense to keep the buttons disabled.

> …
> Removing focusability from disabled elements can offer users both advantages and disadvantages. Allowing keyboard users to skip disabled elements usually reduces the number of key presses required to complete a task. However, preventing focus from moving to disabled elements can hide their presence from screen reader users who "see" by moving the focus.
> …

I have opted for removing the `disabled` attr by default because it feels like that should be default. But I can imagine that somewhere in the future we might add an option to opt-in to this "hard" disable (check the link for some examples).

tbergq (approved):

PR #3194, opened 2021-09-28
Eslint config: add new `eslint-plugin-testing-library`

mrtnzlml:
The enabled rules should be relatively non-invasive so the dependency
upgrade should be quite easy and without breaking changes.


src/eslint-config-adeira/src/presets/react.js
```diff
@@ -211,6 +212,40 @@ module.exports = ({
     'jsx-a11y/scope': ERROR,
     'jsx-a11y/tabindex-no-positive': ERROR,
   },
+  overrides: [
+    {
+      // React Testing Library (https://github.com/testing-library/eslint-plugin-testing-library)
+      files: ['**/__tests__/**/*.[jt]s?(x)', '**/?(*.)+(spec|test).[jt]s?(x)'],
+      rules: {
+        'testing-library/await-async-query': OFF, // TODO: seems to be broken for some `findAllByProps` in our codebase
```
tbergq, line 220: Interesting, did you create an issue on their github?

src/eslint-config-adeira/src/presets/react.js
```diff
@@ -211,6 +212,40 @@ module.exports = ({
...
+        'testing-library/no-await-sync-events': OFF,
+        'testing-library/no-await-sync-query': NEXT_VERSION_ERROR,
+        'testing-library/no-container': OFF,
+        'testing-library/no-debug': NEXT_VERSION_ERROR,
+        'testing-library/no-dom-import': [NEXT_VERSION_ERROR, 'react'],
+        'testing-library/no-manual-cleanup': OFF,
+        'testing-library/no-node-access': OFF,
+        'testing-library/no-promise-in-fire-event': NEXT_VERSION_ERROR,
+        'testing-library/no-render-in-setup': OFF,
+        'testing-library/no-unnecessary-act': OFF,
```
tbergq, line 233: Wouldn't this be of help?

tbergq (approved):

src/eslint-config-adeira/src/presets/react.js
mrtnzlml: It would. I changed it to `NEXT_VERSION_ERROR`. The thing is that there are many other rules from `testing-library` that should be enabled but my thinking is to do it gradually to ease the migration. I am also planning to include https://github.com/testing-library/eslint-plugin-jest-dom, but again: better to do it gradually in the next release for example. Hopefully, that makes sense.

src/eslint-config-adeira/src/presets/react.js
mrtnzlml: I did not. I didn't have much time for it but most importantly it was not very obvious to me what's the problem and if it's even a problem with `testing-library`. I thought it's going to be better to move forward and disable this for now and return to it later (since I am going to be enabling more rules and improving the plugin setup in general).

PR #3195, opened 2021-09-28
SX Design: improve source strings for translations

mrtnzlml:
Specifically, these strings made it hard to understand what needs to be
translated and how.


src/sx-design/src/Modal/ModalDialog.js
```diff
@@ -38,7 +38,9 @@ export default function ModalDialog(props: Props): Node {
                 size="small"
                 onClick={props.onClose}
                 aria-label={
-                  <fbt desc="close modal button ARIA description">close modal button</fbt>
+                  <fbt desc="ARIA description of the button which closes the modal dialog">
+                    button for closing the modal window
```
tbergq, line 42: See #3201, it shouldn't be necessary to have `button` in the aria-label text, as the screen-reader will announce that it is a button.

mrtnzlml: Makes sense! 👍

tbergq (approved):

PR #3202, opened 2021-09-29
SX Design: rotate chevron icon in Breadcrumb

tbergq:
Rotate chevron icon for rtl rather than rendering a different icon.
Pass setLocale function into story to be able to change locale in
stories.


tbergq:
@mrtnzlml the PR should only contain the rotating icon now.

mrtnzlml:
> Ahh, nice, I'll revert that change, I didn't know about this in storybook 👌

It's non-standard, I added it there. Will try to find a way how to document it better.

mrtnzlml (approved):

PR #3204, opened 2021-09-29
example-relay: add next bundle analyzer

tbergq:


src/example-relay/next.config.js
```diff
@@ -3,35 +3,40 @@
...
-      __dirname,
-      // @x-shipit-enable: 'babel.config.js',
-      '.babelrc.js', // @x-shipit-disable
-    ),
-    webpack: (config) => {
-      config.module.rules.push({
-        type: 'javascript/auto',
-        test: /\.mjs$/,
-      });
+module.exports = (withBundleAnalyzer(
```
mrtnzlml, line 15: Tip: there is a great `next-compose-plugins` that makes the Next.js configuration much more approachable, see: https://github.com/adeira/universe/blob/c96ff38e23de5eb9662ae7a3fdf160aa1c17d8c6/src/abacus-backoffice/next.config.js

mrtnzlml (approved):

src/example-relay/next.config.js
tbergq: Nice, I was thinking something like that would be nice. I'll add a separate PR for it.

PR #3213, opened 2021-09-30
fetch: use real timers

### LINKED ISSUE #2436: Jest: migrate from legacy timers
We are currently using `legacy` Jest timers (https://jestjs.io/docs/configuration#timers-string) and we should migrate them.

tbergq:
closes #2436

### DIFF
--- a/flow-typed/msw.js
+++ b/flow-typed/msw.js
@@ -1,0 +1,33 @@
+// @flow
+
+declare module 'msw' {
+  declare type Context = {
+    status: (number) => void,
+    delay: (number) => void,
+    json: (any) => void,
+    ...
+  };
+
+  declare type RestMethod = (
+    url: string,
+    (req: any, res: any, ctx: Context) => Promise<any>,
+  ) => void;
+  declare type Rest = {
+    get: RestMethod,
+    put: RestMethod,
+    post: RestMethod,
+    delete: RestMethod,
+  };
+  declare export var rest: Rest;
+}
+
+declare module 'msw/node' {
+  declare export type Server = {
+    listen: (config: { [key: string]: mixed }) => Promise<void>,
+    resetHandlers: () => Promise<void>,
+    close: () => Promise<void>,
+    use: (any) => void,
+    ...
+  };
+  declare export function setupServer(): Server;
+}
--- a/src/fetch/scripts/setupJest.js
+++ b/src/fetch/scripts/setupJest.js
@@ -1,0 +1,14 @@
+// @flow
+
+import { server } from '../src/test-utils';
+
+beforeAll(() =>
+  server.listen({
+    onUnhandledRequest: 'warn',
+  }),
+);
+// if you need to add a handler after calling setupServer for some specific test
+// this will remove that handler for the rest of them
+// (which is important for test isolation):
+afterEach(() => server.resetHandlers());
+afterAll(() => server.close());
--- a/src/fetch/src/__tests__/fetchWithRetries.403.test.js
+++ b/src/fetch/src/__tests__/fetchWithRetries.403.test.js
@@ -1,40 +1,17 @@
 // @flow
 
-import fetch from '../fetch';
 import fetchWithRetries from '../fetchWithRetries';
-import flushPromises from './_flushPromises';
-
-jest.mock('../fetch');
-
-beforeEach(() => {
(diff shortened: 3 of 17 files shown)


flow-typed/msw.js
```diff
@@ -0,0 +1,33 @@
+// @flow
```
tbergq, line 1: This probably needs a lot of improvements, but it is a start

src/fetch/src/__tests__/fetchWithRetries.403.test.js
```diff
@@ -1,40 +1,17 @@
...
+  server.use(
+    rest.get(url, (_, res, ctx) => {
+      return res(ctx.status(403));
+    }),
+  );
 
-  expect(handleNext).not.toBeCalled();
-  expect(handleCatch).toBeCalledWith(
+  await expect(fetchWithRetries(url, {})).rejects.toEqual(
     new Error('fetchWithRetries: Still no successful response after 1 retries, giving up.'),
```
tbergq, line 15: Technically, it is 0 retries, and 1 attempt, but 🤷

src/fetch/src/__tests__/fetchWithRetries.retries.test.js
```diff
@@ -1,37 +1,29 @@
...
-  expect(fetch.mock.calls).toHaveLength(retries + 1);
-
-  // Timeout last request.
-  await flushPromises();
-  jest.runAllTimers();
+  await expect(fetchWithRetries(url, { retryDelays, fetchTimeout: 100 })).rejects.toThrow(
+    'fetchWithRetries: Failed to get response from server (https://localhost), tried 4 times.',
+  );
+  await (() => {
+    // We have to wait for the last promise to resolve, if now we have an open handle
```
mrtnzlml, line 22: Probably:

```suggestion
    // We have to wait for the last promise to resolve, if not we have an open handle
```

src/fetch/scripts/setupJest.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+
+import { server } from '../src/test-utils';
+
+beforeAll(() =>
+  server.listen({
+    onUnhandledRequest: 'warn',
```
mrtnzlml, line 7: Would `error` or some other kind of hard fail mechanism make more sense here?

flow-typed/msw.js
mrtnzlml: Can you please add there a to-do comment to move it to `flow-typed` repository where it belongs?

mrtnzlml (approved):
Very different kind of testing but seems to be more approachable. 👍

mrtnzlml:
Ay no, it got merged already. 😞 

@tbergq careful with the `automerge` label when the PR is bigger and might receive some comments.

src/fetch/scripts/setupJest.js
tbergq: Yeah, I thought so as well, but it doesn't really work that way. It is just about which console method you want to use, and since we have our test fail on unwanted console logs, this should be good enough.

tbergq:
> Ay no, it got merged already. 😬
> 
> @tbergq careful with the `automerge` label when the PR is bigger and might receive some comments.

Yeah, I didn't expect an approval if you wanted some changes ;) I'll use it with care.

tbergq:
> Very different kind of testing but seems to be more approachable. 👍

Yeah, I prefer `msw` to just mocking fetch for this reason. We also got rid of the weird timeout error messages we had, and are now getting the expected error from our tests 🙂

mrtnzlml:
> > Ay no, it got merged already. 😬
> 
> > 
> 
> > @tbergq careful with the `automerge` label when the PR is bigger and might receive some comments.
> 
> 
> 
> Yeah, I didn't expect an approval if you wanted some changes 😉  I'll use it with care. 

My bad as well. 🙈

src/fetch/scripts/setupJest.js
mrtnzlml: Ah ok. I checked it only briefly and I saw there some callback option as well. So I thought we could throw there. But I didn't spend much time reading it. 😬

tbergq: I thought so as well, and I tried doing so once, but it didn't fail the test.

PR #3214, opened 2021-10-01
SX Design: clarify Storybook languages and themes switching

mrtnzlml:


tbergq (approved):
Thanks, this is good. I would consider adding a code comment in `preview.js` as well. Something like, `locale and theme can be changed in the UI, see readme for more details`.

tbergq (approved):

PR #3539, opened 2021-12-06
SX Design: fix `Money` and `MoneyFn` rendering

mrtnzlml:
Something changed between Node.js 17.1 and 17.2 and our previous
implementation stopped working. Regardless of this fact, it seems like
we should've used numbering system `latn` in a first place.

This commit fixes the issue and unblocks all pending pull requests.


mrtnzlml:
cc @adeira/devs

michalsanger (approved):

PR #3625, opened 2021-12-18
Eslint config: convert all `NEXT_VERSION_ERROR` to `ERROR`

mrtnzlml:
In preparation for next major version.


tbergquist-godaddy (approved):
There might be one more rule to enable here because of #3631

PR #3741, opened 2022-01-12
CI: run Relay check

mrtnzlml:
This makes sure that our generated files are always up-to-date. This
check works for the whole monorepo regardless of the projects inside
(all Relay projects will be checked).


mrtnzlml:
This was previously annoying to do in monorepos but the new Rust Compiler is a game-changer. 😊

tbergquist-godaddy (approved):

tbergquist-godaddy (approved):

PR #3960, opened 2022-02-17
Graphql Global ID: extend graphql peer dependency

michalsanger:


src/graphql-global-id/package.json
```diff
@@ -19,6 +19,6 @@
     "@babel/runtime": "^7.17.2"
   },
   "peerDependencies": {
-    "graphql": "^15.0.0"
+    "graphql": "^16.0.0"
```
mrtnzlml, line 22: What about something like `^15.0.0 || ^16.0.0` to unblock it for version 16? There are still some projects in Universe that depend on this package and they are not upgraded to the latest GraphQL version.

mrtnzlml (approved):
Feel free to release a new version (either in this PR directly or follow-up PR - up to you).

PR #4002, opened 2022-02-24
Abacus: expand permissions for employees

mrtnzlml:


mrtnzlml:
Interesting! So it's broken in runtime but not in tests. Fixing runtime breaks tests. 🤔

PR #4039, opened 2022-03-04
SX: implement styles rehydration

mrtnzlml:
TKTK


src/sx/src/injectRuntimeStyles.js
```diff
@@ -73,42 +44,41 @@ export function injectRuntimeKeyframes(css: string, name: string) {
...
+
+        // Apply the missing styles:
         styleSheet.insertRule(rule, insertIndex);
+
+        // Register the rule into rehydrated styles, so we don't attempt to process it again:
+        rehydratedStyleRules.add(`.${node.getHash()}`);
       }
     } else if (node instanceof StyleCollectorAtNode) {
-      // TODO: make sure we are not adding already added styles (?)
+      // TODO: leverage rehydrated rules (to add only new rules)
```
mrtnzlml, line 72: There is still one missing piece -> it's difficult to convert between `StyleCollectorAtNode` and the rehydrated style identifier. I have this change WIP locally as well.

PR #4047, opened 2022-03-05
Configure Renovate bot

mrtnzlml:
History repeats: I am configuring Renovate bot again. Dependabot is
great and it's awesome that it's integrated into GitHub natively.
Unfortunately, it does not support Yarn Berry and seems like it won't
anytime soon (https://github.com/dependabot/dependabot-core/issues/1297).

I would like to configure Renovate bot instead which should support Yarn
Berry without many problems and therefore unblock our Yarn upgrade (https://github.com/adeira/universe/pull/2154).

At this moment I would like to keep both Dependabot and Renovate bot
enabled to see if everything works correctly, and eventually I'd like to
turn off Dependabot completely or limit it to things Renovate bot doesn't
support.

Closes: https://github.com/adeira/universe/pull/3799


.github/renovate.json5
```diff
@@ -0,0 +1,14 @@
+{
```
mrtnzlml, line 1: This should be somehow similar to what we do with Dependabot: https://github.com/adeira/universe/blob/3b3e1e4bca7a63262b1dcba2187be9e003dbeb6a/.github/dependabot.yml

tbergquist-godaddy (approved):

PR #4049, opened 2022-03-06
Workspaces scan: flip check for `private` packages

mrtnzlml:
In Yarn Berry the package is _never_ going to be `private:false` but it
can be `private:true`. This commit fixes the condition.


mrtnzlml:
This PR also offloads https://github.com/adeira/universe/pull/2154

tbergquist-godaddy (approved):

PR #4090, opened 2022-03-07
Yarn: install `interactive-tools` plugin

mrtnzlml:
See: https://github.com/yarnpkg/berry/blob/HEAD/packages/plugin-interactive-tools/README.md


mrtnzlml:
https://yarnpkg.com/cli/upgrade-interactive

PR #4194, opened 2022-03-31
Use React 18

mrtnzlml:
See: https://reactjs.org/blog/2022/03/29/react-v18.html


mrtnzlml:
One of the failing tests is `src/sx-design/src/Button/__tests__/Button.test.js`. This led me to `Icon` component as being the culprit. 🤔

PR #4391, opened 2022-05-10
SX Design: patch 3rd party dependency

mrtnzlml:
See: https://github.com/storybookjs/storybook/issues/18019

Should we resolved in `ip` version 1.1.6 or 2.0.0, see: https://github.com/indutny/node-ip/commit/4de50aec875d12b004849e11e19d6daf68b50c2d


.yarn/patches/ip-npm-1.1.5-af36318aa6.patch
```diff
@@ -0,0 +1,658 @@
+diff --git a/.idea/.gitignore b/.idea/.gitignore
+new file mode 100644
+index 0000000000000000000000000000000000000000..b58b603fea78041071d125a30db58d79b3d49217
+--- /dev/null
++++ b/.idea/.gitignore
```
mrtnzlml, line 5: This is weird, but... 🤷

PR #4392, opened 2022-05-10
SX Design: test Storybook via Playwright

mrtnzlml:
Just to make sure it still works after updating Storybook dependencies as it gets broken _very_ often.


mrtnzlml:
The CI fail should be related to: https://github.com/storybookjs/storybook/issues/18019

PR #4448, opened 2022-05-22
SX Design: init visual tests (`Kbd`)

mrtnzlml:


mrtnzlml:
Mm, this situation is currently not the best, see: https://github.com/microsoft/playwright/discussions/12729

Better no.

PR #5043, opened 2022-09-17
Relay: enforce correct type for `id` fields

mrtnzlml:
See: https://github.com/facebook/relay/commit/d0a39283d45854922119b779c4ac7014fd402d20


mrtnzlml:
Cc @KrysKruk 😎

Issue #5188, opened 2022-10-19
CI: install and use `shellcheck`


mrtnzlml:
See: https://github.com/koalaman/shellcheck

Other than regular `sh` scripts we also need to check `./x`

PR #5207, opened 2022-10-22
CI: add `shellcheck`

mrtnzlml:
Closes: https://github.com/adeira/universe/issues/5188


x
```diff
@@ -41,7 +41,7 @@
 # ./x r -- --help
 # ```
 
-if ! command -v rustup &> /dev/null ; then
+if!command -v rustup &> /dev/null ; then
```
mrtnzlml, line 44: TODO: revert

PR #6112, opened 2023-05-18
CI: run JS tests on `macos-latest`

mrtnzlml:
I suspect that one test is broken on macOS. This is just a test. DO NOT MERGE.


mrtnzlml:
This is strange. It works here on GitHub (which is great because this is where we are executing Shipit from), but it's broken on both of my MacBooks (M1 as well as pre-M1). Not sure what is going on. 🤔

PR #6138, opened 2023-05-23
Fixes some edge cases in monorepo-shipit

itsdouges:
This pull request:

- fixes hardcoded branches during the shipit verify step
- includes the first revision commit when running in a new repo
- adds `--committer-date-is-author-date` to `git am` call in `commitPatch` to ensure each commit uses the author datetime, preventing the initial sync forcing every commit to have the same commit timestamp 

With these changes (applied as a patch) I've been able to sync a repo in CI successfully.

### DIFF
--- a/src/monorepo-shipit/src/RepoGit.js
+++ b/src/monorepo-shipit/src/RepoGit.js
@@ -19,6 +19,7 @@ export interface SourceRepo {
     baseRevision: string,
     headRevision: string,
     roots: Set<string>,
+    includeBaseRevision?: boolean,
   ): null | $ReadOnlyArray<string>;
   getChangesetFromID(revision: string): Changeset;
   getNativePatchFromID(revision: string): string;
@@ -213,7 +214,8 @@ export default class RepoGit implements AnyRepo, SourceRepo, DestinationRepo {
     baseRevision: string,
     headRevision: string,
     roots: Set<string>,
-  ) => null | $ReadOnlyArray<string> = (baseRevision, headRevision, roots) => {
+    includeBaseRevision?: boolean,
+  ) => null | $ReadOnlyArray<string> = (baseRevision, headRevision, roots, includeBaseRevision) => {
     const log = this._gitCommand(
       'log',
       '--reverse',
@@ -227,7 +229,17 @@ export default class RepoGit implements AnyRepo, SourceRepo, DestinationRepo {
       .runSynchronously()
       .getStdout();
     const trimmedLog = log.trim();
-    return trimmedLog === '' ? null : trimmedLog.split('\n');
+    if (trimmedLog === '') {
+      return null;
+    }
+
+    const descendants = trimmedLog.split('\n');
+
+    if (includeBaseRevision === true) {
+      descendants.unshift(baseRevision);
+    }
+
+    return descendants;
   };
 
   commitPatch: (changeset: Changeset) => string = (changeset) => {
--- a/src/monorepo-shipit/src/RepoGit.js
+++ b/src/monorepo-shipit/src/RepoGit.js
@@ -258,7 +258,9 @@ export default class RepoGit implements AnyRepo, SourceRepo, DestinationRepo {
     } else {
       const diff = this.renderPatch(changeset);
       try {
-        this._gitCommand('am', '--keep-non-patch', '--keep-cr').setStdin(diff).runSynchronously();
+        this._gitCommand('am', '--keep-non-patch', '--keep-cr', '--committer-date-is-author-date')
+          .setStdin(diff)
+          .runSynchronously();
       } catch (error) {
         this._gitCommand('am', '--show-current-patch').setOutputToScreen().runSynchronously();
         this._gitCommand('am', '--abort').setOutputToScreen().runSynchronously();
--- a/src/monorepo-shipit/src/phases/createSyncPhase.js
+++ b/src/monorepo-shipit/src/phases/createSyncPhase.js
@@ -18,16 +18,20 @@ export default function createSyncPhase(config: ShipitConfig): Phase {
     const destinationRepo = _getDestinationRepo();
     const sourceRepo = _getSourceRepo();
     let initialRevision = destinationRepo.findLastSourceCommit(config.getDestinationRoots());
+    let isNewRepo = false;
     if (initialRevision === null) {
       // Seems like it's a new repo so there is no signed commit.
       // Let's take the first one from our source repo instead.
       initialRevision = sourceRepo.findFirstAvailableCommit();
+      isNewRepo = true;
     }
+
     const sourceChangesets = new Set<Changeset>();
     const descendantsPath = sourceRepo.findDescendantsPath(
       initialRevision,
       config.getSourceBranch(),
       config.getSourceRoots(),
+      isNewRepo,
     );
     if (descendantsPath !== null) {
       descendantsPath.forEach((revision) => {
--- a/src/monorepo-shipit/src/phases/createVerifyRepoPhase.js
+++ b/src/monorepo-shipit/src/phases/createVerifyRepoPhase.js
@@ -82,7 +82,7 @@ export default function createVerifyRepoPhase(config: ShipitConfig): Phase {
       'diff',
       '--stat',
       'HEAD',
-      'shipit_destination/master',
+      `shipit_destination/${config.getDestinationBranch()}`,
     )
       .runSynchronously()
       .getStdout()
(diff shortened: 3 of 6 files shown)


src/monorepo-shipit/src/phases/createSyncPhase.js
```diff
@@ -18,16 +18,20 @@ export default function createSyncPhase(config: ShipitConfig): Phase {
     const destinationRepo = _getDestinationRepo();
     const sourceRepo = _getSourceRepo();
     let initialRevision = destinationRepo.findLastSourceCommit(config.getDestinationRoots());
+    let isNewRepo = false;
     if (initialRevision === null) {
       // Seems like it's a new repo so there is no signed commit.
       // Let's take the first one from our source repo instead.
       initialRevision = sourceRepo.findFirstAvailableCommit();
+      isNewRepo = true;
```
itsdouges, line 26: We assume it's a new repo when this branch is executed.

src/monorepo-shipit/src/phases/createSyncPhase.js
```diff
@@ -18,16 +18,20 @@ export default function createSyncPhase(config: ShipitConfig): Phase {
...
     }
+
     const sourceChangesets = new Set<Changeset>();
     const descendantsPath = sourceRepo.findDescendantsPath(
       initialRevision,
       config.getSourceBranch(),
       config.getSourceRoots(),
+      isNewRepo,
     );
```
itsdouges, line 35: Pass it to the descendants function.

src/monorepo-shipit/src/RepoGit.js
```diff
@@ -227,7 +229,17 @@ export default class RepoGit implements AnyRepo, SourceRepo, DestinationRepo {
...
     const trimmedLog = log.trim();
-    return trimmedLog === '' ? null : trimmedLog.split('\n');
+    if (trimmedLog === '') {
+      return null;
+    }
+
+    const descendants = trimmedLog.split('\n');
+
+    if (includeBaseRevision === true) {
+      descendants.unshift(baseRevision);
```
itsdouges, line 239: Since the above command doesn't include the base revision we conditionally add it at the beginning here.

src/monorepo-shipit/src/RepoGit.js
```diff
@@ -110,13 +111,7 @@ export default class RepoGit implements AnyRepo, SourceRepo, DestinationRepo {
...
 
   findLastSourceCommit: (roots: Set<string>) => null | string = (roots) => {
-    const log = this._gitCommand(
-      'log',
-      '-1',
-      '--grep',
-      '^adeira-source-id: \\?[a-z0-9]\\+\\s*$',
-      ...roots,
-    )
+    const log = this._gitCommand('log', '-1', '--grep', '^adeira-source-id: [a-z0-9]*$', ...roots)
```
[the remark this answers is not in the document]
itsdouges: This causes one unit test to fail. Oddly enough if I revert it and run the tests locally they all fail. Is this a problem with bash vs. zsh. vs. ?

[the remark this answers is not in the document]
itsdouges: For now I've reverted with the hope that it works fine in CI just not during my local. I'm going to make a git patch to find out.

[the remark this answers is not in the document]
itsdouges: Yes that seems to be the case. Odd!

image

src/monorepo-shipit/src/RepoGit.js
```diff
@@ -246,7 +258,9 @@ export default class RepoGit implements AnyRepo, SourceRepo, DestinationRepo {
     } else {
       const diff = this.renderPatch(changeset);
       try {
-        this._gitCommand('am', '--keep-non-patch', '--keep-cr').setStdin(diff).runSynchronously();
+        this._gitCommand('am', '--keep-non-patch', '--keep-cr', '--committer-date-is-author-date')
```
itsdouges, line 261: `--committer-date-is-author-date` ensures that the commit date is the same as the author date (instead of the timestamp of the commit). This results in a nicer commit history when syncing a large repo for the first time, instead of everything being the same commit timestamp they're the appropriate:

image

itsdouges: And here is a subsequent run, works great: 

Source:
image

Destination:
image

mrtnzlml: This seems to be a great addition; thanks! 👍

src/monorepo-shipit/src/RepoGit.js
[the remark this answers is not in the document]
mrtnzlml: @itsdouges I also had the failing test locally, but not anymore. I described the issue here: https://github.com/adeira/universe/pull/6114 (second half of the description)

TL;DR: `brew install git` because you _probably_ have the older broken version of Git on macOS.

src/monorepo-shipit/src/RepoGit.js
mrtnzlml: I don't understand this change. Can you please explain it better?

I've used Shipit to export into new repositories many times, and it always worked well. Maybe the problem is somewhere else? 🤔

itsdouges: So my initial commit on my monorepo isn't empty. Without this code this function never returns the first commit, but only the second. This means when changes are applied it starts with the second commit, and the diff is applying (via git am) to files that don't yet exist (because the git am application from the first commit hasn't happened!).

To get around this, without having to re-write my monorepos history to add an empty commit at the beginning of time, we can conditionally include the first commit (baseRevision) in the descendants and get around the problem.

Does that make sense?

src/monorepo-shipit/src/RepoGit.js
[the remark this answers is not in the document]
itsdouges: Thank you!

src/monorepo-shipit/src/RepoGit.js
itsdouges: To be specific: initial commits are skipped by the `[findDescendantsPath](https://github.com/adeira/shipit/blob/master/src/RepoGit.js#L220-L223)` git command. This is because git log --ancestory-path A..B doesn't include the "A" commit. This is a problem when the initial commit has content like my repo does.

An alternative could be to instead get every commit if it's fresh destination repo? Both would be equivalent.

itsdouges: If you're interested we could catch up on Discord and I can walk you through my monorepo and needs I have from this and beyond? 🙏

mrtnzlml:
Hi @itsdouges, I just wanted to acknowledge this PR, but please be patient - I am swamped at work. 😅

itsdouges:
No worries!

src/monorepo-shipit/src/RepoGit.js
mrtnzlml: @itsdouges Would you please share with me the exact steps to reproduce? What did you do exactly? I am still a bit lost and I would like to try it myself.

Shipit starts exporting from the beginning of the monorepo history if the child repo is either empty or doesn't contain the `adeira-source-id` handle. I've done both many times (export to empty child repo as well as connect Shipit to already existing child repo) and I know it works. Perhaps I just need to explain the process better?

itsdouges: But that's the thing it doesn't do it from the first commit but the second. Let me find a way to reproduce in a unit test and I'll push it up.

itsdouges: I pushed a test. Does that explain things?

itsdouges:
Hey @mrtnzlml would love some eyes on this when you're available - I can separate this into two pull requests if needed (separating out the `includes the first revision commit when running in a new repo` change from everything else.

src/monorepo-shipit/src/RepoGit.js
mrtnzlml: I appreciate your patience and especially the test case - that not only helped me to understand the issue but it also guarantees I won't break this by accident in the future. Thanks! All clear now!

src/monorepo-shipit/src/phases/createVerifyRepoPhase.js
```diff
@@ -82,7 +82,7 @@ export default function createVerifyRepoPhase(config: ShipitConfig): Phase {
       'diff',
       '--stat',
       'HEAD',
-      'shipit_destination/master',
+      `shipit_destination/${config.getDestinationBranch()}`,
```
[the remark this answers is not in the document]
mrtnzlml: Makes sense. 👍

mrtnzlml (approved):

mrtnzlml:
> I can separate this into two pull requests if needed

@itsdouges For future reference - I'd recommend splitting it into multiple PRs if possible. I can review small and isolated PRs much faster.

Anyway, good job. Thank you very much! 👍

PR #6147, opened 2023-05-25
feat: adds commit message transformer to shipit

itsdouges:
This pull request:

- adds the `transformCommitMessage` option to shipit config so consumers can optionally transform their commit messages, such as removing github issues and pull requests from messages

For context currently when running ShipIt it doesn't do anything with commit messages. This is a problem when it exposes internal issues & pull requests that don't exist in the public repo! So, hence this feature add. I can remove them on my end.

### DIFF
--- a/src/monorepo-shipit/ConfigType.flow.js
+++ b/src/monorepo-shipit/ConfigType.flow.js
@@ -10,4 +10,5 @@ export type ConfigType = {
     +source: string,
     +destination: string,
   },
+  +transformCommitMessage?: (message: string) => string,
 };
--- a/src/monorepo-shipit/README.md
+++ b/src/monorepo-shipit/README.md
@@ -98,6 +98,21 @@ module.exports = {
 };
 ```
 
+```js
+export type ConfigType = {
+  +getStaticConfig: () => {
+    +repository: string,
+  },
+  +getPathMappings: () => Map<string, string>,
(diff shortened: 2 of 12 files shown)


itsdouges:
Hey @mrtnzlml would love some eyes on this when you're available - I need this feature before turning on ShipIt for prime time 😃.

mrtnzlml (approved):
@itsdouges I am going to accept this change but I think in the future this is a path I would not like to take. Instead, I'd like to make the filters configurable from outside so that you can pass your custom filters (such as `src/monorepo-shipit/src/filters/stripDescriptions.js`) from the config file.

So instead of having `transformCommitMessage` (which has very specific use-case), you'd have `customFilters` or something similar that would allow you to define any behavior (not just the commit message transformations).

itsdouges:
> @itsdouges I am going to accept this change but I think in the future this is a path I would not like to take. Instead, I'd like to make the filters configurable from outside so that you can pass your custom filters (such as `src/monorepo-shipit/src/filters/stripDescriptions.js`) from the config file.
> 
> 
> 
> So instead of having `transformCommitMessage` (which has very specific use-case), you'd have `customFilters` or something similar that would allow you to define any behavior (not just the commit message transformations).

That makes sense. For the next one that comes up we can work towards it. Fbshipit has this behaviour yeah, where you define your own config (instead of using a default one).

PR #6161, opened 2023-05-27
chore: allow commit transform

itsdouges:
This pull request:

- adds `transformCommitMessage` to allow config values

**TODO**

- [ ] Agree on path forward (see comment)

### DIFF
--- a/src/monorepo-shipit/src/requireAndValidateConfig.js
+++ b/src/monorepo-shipit/src/requireAndValidateConfig.js
@@ -45,6 +45,7 @@ export default function requireAndValidateConfig(configFile: string): ConfigType
     ['getStaticConfig', true],
     ['getPathMappings', true],
     ['getStrippedFiles', false],
+    ['transformCommitMessage', false],
   ]);
 
   // TODO: consider Ajv but with good error messages!


itsdouges:
@mrtnzlml So it looks like I need to also transform the changeset _subject_ property so the git commit title is transformed. Do you have a desired path forward instead of relying on the hardcoded description transform? How would you like to expose custom filters?

Something like this where it gets added onto the end (or wherever appropriate in the pipeline) would be ideal: 

```
getCustomFilters() {
  return [
    stripGithubLinks,
  ];
}
```

Let me know your thoughts and we can unwind the current implementation because it doesn't actually work currently.

mrtnzlml (approved):
Next time, please consider adding tests. 😎

mrtnzlml:
> @mrtnzlml So it looks like I need to also transform the changeset _subject_ property so the git commit title is transformed. Do you have a desired path forward instead of relying on the hardcoded description transform? How would you like to expose custom filters?
> 
> Something like this where it gets added onto the end (or wherever appropriate in the pipeline) would be ideal in the userland config:
> 
> ```
> getCustomFilters() {
>   return {
>     shipit: [stripGithubLinks],
>     // importit as well
>   };
> }
> ```
> 
> Let me know your thoughts and we can unwind the current implementation because it doesn't actually work currently.
> 
> FYI this is the repo I'm syncing too! https://github.com/try-triplex/triplex Works great.

@itsdouges I would probably consider taking an inspiration from `ShipitConfig.getDefaultShipitFilter`, so that would be something like:

```js
type ChangesetFilter = (Changeset) => Changeset;

getCustomShipitFilter(): ChangesetFilter {
  return (changeset: Changeset): Changeset => {
    // do whatever you want here, return modified `Changeset`
  }
}
```

Then we could modify the `getDefaultShipitFilter` like so:

```js
getDefaultShipitFilter(): ChangesetFilter {
  return (changeset: Changeset) => {
    const modifiedChangeset = this.getCustomShipitFilter(changeset);

    // rest of the current implementation ...
  };
}
```

I would also consider renaming `getDefaultShipitFilter` to simply `getShipitFilter`.

What do you think?

itsdouges:
@mrtnzlml that looks good, the rename makes sense. I can add this and remove the commit message transform, the current implementation doesn't update changeset subject only description (my mistake) so it's not fit for purpose atm.

PR #6164, opened 2023-05-27
Shipit: improve CLI interface

mrtnzlml:
The previous solution was good enough when the tool was used internally but now I see the need to have a better CLI interface. This not only provides better out-of-the box validations but also supports `--help` option.

I've also improved the Importit interface to support `--pull-request` option (instead of repo URL and PR number). I believe it's more natural to just copy the PR URL and use it.

### DIFF
--- a/src/monorepo-shipit/bin/importit.js
+++ b/src/monorepo-shipit/bin/importit.js
@@ -3,7 +3,7 @@
 // @flow
 
 import { invariant } from '@adeira/js';
-import commandLineArgs from 'command-line-args';
+import { program as commander } from 'commander';
 
 import iterateConfigs from '../src/iterateConfigs';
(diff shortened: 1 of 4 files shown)


mrtnzlml:
What do you think @itsdouges?

src/monorepo-shipit/bin/importit.js
```diff
@@ -13,79 +13,62 @@ import createImportSyncPhase from '../src/phases/createImportSyncPhase';
...
-    },
-    {
-      name: 'repo-url',
-      type: String,
-    },
-  ],
-  { camelCase: true },
-);
+commander
+  .version(require('./../package.json').version)
```
itsdouges, line 28: nit: would be great to add a description to both shipit and importit

itsdouges:
Nice! The pull request URL change is great as well.

PR #6405, opened 2023-07-11
Abacus tools: add simple API endpoint

mrtnzlml:


mrtnzlml:
TODO: make Playwright tests required for Abacus Tools

PR #6785, opened 2023-10-11
Rossum SA ext: use new `data-sa-extension-schema-id` attribute

mrtnzlml:


mrtnzlml:
Todo: merge when the product change is approved and merged.

PR #6965, opened 2023-11-23
Limit UI improvements to relevant paths & Sort extensions list

miroslav-rehounek:
One caveat is that if the number of extensions per page is smaller than their total (on other pages), it will only sort alphabetically those displayed on that page.

### DIFF
--- a/src/rossum-sa-extension/scripts/list-selector-improvements.js
+++ b/src/rossum-sa-extension/scripts/list-selector-improvements.js
@@ -1,37 +1,66 @@
 // @flow
 
-const listSelectorImprovements = document.createElement('style');
-
-// TODO: apply only on /settings/extensions/logs (?)
-listSelectorImprovements.textContent = `
-  div:has(> ul.MuiMenu-list) {
-    min-width: 400px !important;
-  }`;
-
-document.head?.appendChild(listSelectorImprovements);
+function findAddedDOMElement(mutation, querySelector) {
+  if (mutation.type === 'childList' && mutation.addedNodes.length > 0) {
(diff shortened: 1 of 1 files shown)


src/rossum-sa-extension/scripts/list-selector-improvements.js
```diff
@@ -1,37 +1,66 @@
...
   }
 });
 
 const body = document.body;
 if (body) {
-  listSelectorObserver.observe(body, {
-    childList: true,
-    subtree: true,
-  });
+  if (window.location.pathname.endsWith('/extensions/logs')) {
```
mrtnzlml, line 43: I am afraid this will only work when you refresh the page, and you are already on this URL. However, if you open the page on another URL and transition to this path, it will not run. This would need to be done differently (not sure how, quite frankly). Am I missing something important? 🤔

mrtnzlml (changes requested):
The proposed changes would cause the code to not run unless the openned page has the right URL from the start.

> One caveat is that if the number of extensions per page is smaller than their total (on other pages), it will only sort alphabetically those displayed on that page.

🤔 This is a good point. Maybe we need to reconsider this feature completely. I was thinking that as a prerequisite we could always redirect to a `?pageSize=100` page.

mrtnzlml:
I will close this PR as it doesn't seem to be in the right direction. In fact, I will try to remove the parts of the plugin that are trying to change/fix the FE UI altogether (which means convincing FE to implement these changes into the product itself 🤞).

Thanks for the changes anyway! Much appreciated!

PR #7197, opened 2024-01-15
Rossum Sheets: add `RossumPlugin`

mrtnzlml:
This is quite a big rework that removes the magic parts and focuses on being more explicit (`=ROSSUM.DP("document_id")` instead of `document_id`).

This might seem like a step back, but it allows for patters such as `=ROSSUM.DP_CONFIDENCE("document_id")` or `=ROSSUM.PAYLOAD("annotation.url")`.


mrtnzlml:
Won't pursue this any further. Use "formula fields" instead.

PR #7455, opened 2024-03-14
Rossum EXT: implement resizeable annotation sidebar

mrtnzlml:


mrtnzlml:
Won't pursue this any further.

PR #7525, opened 2024-04-02
Rossum Sheets: add support for matrix operations

mrtnzlml:


mrtnzlml:
I won't pursue this any further. Use "formula fields" instead.

